PR #7, opened 2024-10-21
fix: allow repeated write to same blob

alanshaw:
This PR allows multiple uploads of the same blob in different spaces.

Ideally we'd signal to the client that an upload is not needed, but this is the quicker solution and would possibly be useful if a misbehaving client sends the data anyway.

It also adds server tests to verify this is allowed and fixes a number of bugs that presented themselves.

resolves https://github.com/storacha/storage/issues/6


pkg/service/blobs/server.go
```diff
@@ -129,10 +116,10 @@ func NewBlobPutHandler(presigner presigner.RequestPresigner, allocs allocationst
 		}
 
 		err = blobs.Put(r.Context(), digest, uint64(contentLength), r.Body)
-		if err == nil {
+		if err != nil {
```
hannahhoward, line 119: curious was this just a bug before?

hannahhoward (approved):
LGTM seems an adequate solution for now.

pkg/service/blobs/server.go
alanshaw: Yep 🤦‍♂️

PR #10, opened 2024-10-21
feat: cache location claim in indexing service

alanshaw:
This PR allows configuration of the indexing service DID/URL and upgrades `publisher.Publish(...)` to make a `claim/cache` invocation to it.

TODO: actually send the claim in the invocation blocks (needs ucanto upgrade). See: https://github.com/storacha/go-ucanto/pull/27

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/storacha/go-ucanto/principal"
 	ed25519 "github.com/storacha/go-ucanto/principal/ed25519/signer"
 	"github.com/storacha/storage/pkg/server"
+	"github.com/storacha/storage/pkg/service/publisher"
 	"github.com/storacha/storage/pkg/service/storage"
 	"github.com/storacha/storage/pkg/store/blobstore"
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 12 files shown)


cmd/main.go
```diff
@@ -55,32 +56,46 @@ func main() {
...
-						}
-						log.Errorf("Server ID is not configured, generated one for you: %s", id.DID().String())
-					} else {
-						id, err = ed25519.Parse(cCtx.String("private-key"))
-						if err != nil {
-							return fmt.Errorf("parsing private key: %w", err)
+
+					pkstr := cCtx.String("private-key")
+					if pkstr == "" {
+						if os.Getenv("PRIVATE_KEY") != "" {
```
hannahhoward, line 62: suggestion: instead of doing this you can just set the EnvVars property on the CLI flag
https://cli.urfave.org/v2/examples/flags/#values-from-the-environment

cmd/main.go
```diff
@@ -55,32 +56,46 @@ func main() {
...
 					if err != nil {
 						return fmt.Errorf("getting user home directory: %w", err)
 					}
 
 					dataDir := cCtx.String("data-dir")
 					if dataDir == "" {
-						dir, err := mkdirp(homeDir, ".storacha")
-						if err != nil {
-							return err
+						if os.Getenv("DATA_DIR") != "" {
```
hannahhoward, line 89: same comment as above.

cmd/main.go
```diff
@@ -115,21 +130,54 @@ func main() {
 
 					pubURLstr := cCtx.String("public-url")
 					if pubURLstr == "" {
-						pubURLstr = fmt.Sprintf("http://localhost:%d", cCtx.Int("port"))
-						log.Errorf("Public URL is not configured, using: %s", pubURLstr)
+						if os.Getenv("PUBLIC_URL") != "" {
```
hannahhoward, line 133: same comment as above.

cmd/main.go
```diff
@@ -115,21 +130,54 @@ func main() {
...
+							log.Errorf("Public URL is not configured, using: %s", pubURLstr)
+						}
 					}
 					pubURL, err := url.Parse(pubURLstr)
 					if err != nil {
 						return fmt.Errorf("parsing public URL: %w", err)
 					}
 
+					announceURL := *publisher.AnnounceURL
+					if os.Getenv("ANNOUNCE_URL") != "" {
```
hannahhoward, line 146: any reason this one and the next two appear to be env only? I would just make them CLI flags too that can read from env (but can also be set at startup)

pkg/service/publisher/publisher.go
```diff
@@ -15,22 +14,36 @@ import (
...
+	"github.com/storacha/storage/pkg/capability"
 	"github.com/storacha/storage/pkg/service/publisher/advertisement"
 )
 
-const claimPattern = "{claim}"
-
 var log = logging.Logger("publisher")
 
+var (
+	AnnounceURL, _        = url.Parse("https://cid.contact/announce")
```
hannahhoward, line 35: I would make these CLI flags, with these as default values (also possible -- use the Value parameter). It's very confusing to pull them out of here and only use them in the CLI. unless they are used somewhere else?

hannahhoward (changes requested):
LGTM but I would clean the use of cli flags, env vars, and default values (into just cli flags that can support both env vars and default values0

cmd/main.go
alanshaw: Oh I had no idea you could do that! Awesome.

pkg/service/publisher/publisher.go
alanshaw: Ok I've moved these into `cmd/main.go`.

PR #14, opened 2024-10-26
feat: map blobstore implements filesystemer

alanshaw:
This PR enables the map blobstore to present itself as a `http.FileSystem`. This is useful for testing with an ephemeral storage node.

### DIFF
--- a/pkg/service/blobs/server.go
+++ b/pkg/service/blobs/server.go
@@ -36,19 +36,7 @@ func NewBlobGetHandler(blobs blobstore.Blobstore) func(http.ResponseWriter, *htt
 	if fsblobs, ok := blobs.(blobstore.FileSystemer); ok {
 		serveHTTP := http.FileServer(fsblobs.FileSystem()).ServeHTTP
 		return func(w http.ResponseWriter, r *http.Request) {
-			_, bytes, err := multibase.Decode(r.PathValue("blob"))
-			if err != nil {
-				http.Error(w, fmt.Sprintf("decoding multibase encoded digest: %s", err), http.StatusBadRequest)
-				return
-			}
-
-			digest, err := multihash.Cast(bytes)
-			if err != nil {
-				http.Error(w, fmt.Sprintf("invalid multihash digest: %s", err), http.StatusBadRequest)
-				return
-			}
-
-			r.URL.Path = fsblobs.EncodePath(digest)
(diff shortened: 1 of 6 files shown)


pkg/service/blobs/server.go
```diff
@@ -36,19 +36,7 @@ func NewBlobGetHandler(blobs blobstore.Blobstore) func(http.ResponseWriter, *htt
 	if fsblobs, ok := blobs.(blobstore.FileSystemer); ok {
 		serveHTTP := http.FileServer(fsblobs.FileSystem()).ServeHTTP
 		return func(w http.ResponseWriter, r *http.Request) {
-			_, bytes, err := multibase.Decode(r.PathValue("blob"))
```
hannahhoward, line 39: this might be simpler with ServeFile instead of FileServer.

also go errata: if you just want to use system utils, you may find ServeFileFS and FileServerFS easer cause you can just use some system utils (if this is actually helpful)

pkg/service/storage/server.go
```diff
@@ -28,7 +28,12 @@ func (srv *Server) Serve(mux *http.ServeMux) {
 
 func NewHandler(server server.ServerView) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
-		res, _ := server.Request(ucanhttp.NewHTTPRequest(r.Body, r.Header))
+		res, err := server.Request(ucanhttp.NewHTTPRequest(r.Body, r.Header))
+		if err != nil {
```
hannahhoward, line 32: +1

pkg/store/blobstore/fs.go
```diff
@@ -194,3 +191,21 @@ func NewFsBlobstore(rootdir string, tmpdir string) (*FsBlobstore, error) {
 	}
 	return &FsBlobstore{rootdir, tmpdir}, nil
 }
+
+type fsDir struct {
```
hannahhoward, line 195: again you may want to look at fs.FS

hannahhoward (approved):
LGTM with some suggestions to look into (non-blocking)

pkg/service/blobs/server.go
alanshaw: Ah yes - looks more appropriate. I didn't realize it existed! If I have a moment I'll look into replacing it.

PR #15, opened 2024-10-28
feat: return location commitment in blob/accept receipt

alanshaw:
This PR augments the blob/accept handler to return the location commitment in the receipt effects.


hannahhoward (approved):
LGTM

PR #16, opened 2024-10-28
fix(publisher): server paths

alanshaw:
This PR fixes the server paths - only return version info when direct hit on `/` and fix for IPNI advert path - includes a CID.


hannahhoward (approved):
LGTM

PR #20, opened 2024-10-30
fix: blob/allocate response for already allocated or received blob

alanshaw:
This PR fixes the response for a repeated `blob/allocate` for an existing blob.

Previously:

* Not allocated and not received - allocated size: {blob size} and signed URL for put
* Not allocated and received (allocated in another space) - allocated size: {blob size} and signed URL for put
* Allocated but not received - allocated size: {blob size} and signed URL for put
* Allocated and received - allocated size: 0, no URL

Now:

* Not allocated and not received - allocated size: {blob size} and signed URL for put
* Not allocated and received (allocated in another space) - allocated size: {blob size}, no URL
* Allocated but not received - allocated size: 0 (already allocated in the space) and signed URL for put
* Allocated and received - allocated size: 0, no URL

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -17,7 +17,7 @@ import (
 
 var log = logging.Logger("server")
 
-// ListenAndServe creates a new indexing service HTTP server, and starts it up.
+// ListenAndServe creates a new storage node HTTP server, and starts it up.
 func ListenAndServe(addr string, service storage.Service) error {
 	srvMux, err := NewServer(service)
 	if err != nil {
--- a/pkg/service/storage/ucan.go
+++ b/pkg/service/storage/ucan.go
@@ -41,7 +41,7 @@ func NewUCANServer(storageService Service) (server.ServerView, error) {
 					if cap.With() != iCtx.ID().DID().String() {
 						return blob.AllocateOk{}, nil, NewUnsupportedCapabilityError(cap)
 					}
-					// check if we already have an allcoation for the blob in this space
+					// check if we already have an allocation for the blob in this space
 					allocs, err := storageService.Blobs().Allocations().List(ctx, digest)
 					if err != nil {
 						log.Errorf("getting allocations: %w", err)
--- a/pkg/service/storage/ucan.go
+++ b/pkg/service/storage/ucan.go
@@ -41,7 +41,6 @@ func NewUCANServer(storageService Service) (server.ServerView, error) {
 					if cap.With() != iCtx.ID().DID().String() {
 						return blob.AllocateOk{}, nil, NewUnsupportedCapabilityError(cap)
 					}
-
 					// check if we already have an allcoation for the blob in this space
 					allocs, err := storageService.Blobs().Allocations().List(ctx, digest)
 					if err != nil {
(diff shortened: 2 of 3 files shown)


pkg/service/storage/ucan.go
```diff
@@ -41,41 +41,72 @@ func NewUCANServer(storageService Service) (server.ServerView, error) {
...
+						}
+						if err != nil && !errors.Is(err, store.ErrNotFound) {
+							log.Errorf("getting blob: %w", err)
+							return blob.AllocateOk{}, nil, failure.FromError(err)
 						}
 					}
 
+					// the size reported in the receipt is the number of bytes allocated
+					// in the space - if a previous allocation already exists, this has
+					// already been done, so the allocation size is 0
```
volmedo, line 74: q: does that imply that only the first space where a given blob was allocated is billed for that storage or this has nothing to do with charging users?

pkg/service/storage/ucan.go
```diff
@@ -41,41 +41,72 @@ func NewUCANServer(storageService Service) (server.ServerView, error) {
...
+						}
+						address = &blob.Address{
+							URL:     url,
+							Headers: headers,
+							Expires: expiresAt,
+						}
 					}
 
+					// even if a previous allocation was made in this space, we create
+					// another for the new invocation.
```
volmedo, line 109: q: this is for the case that a previous allocation invocation expired, right?

pkg/service/storage/ucan_test.go
```diff
@@ -132,6 +132,78 @@ func TestServer(t *testing.T) {
...
+			fmt.Printf("%+v\n", ok)
+			require.Equal(t, int64(0), ok.Size)
+			require.Nil(t, ok.Address)
+		}, func(x ipld.Node) {
+			f := testutil.BindFailure(t, x)
+			fmt.Println(f.Message)
+			fmt.Println(*f.Stack)
+			require.Nil(t, f)
+		})
+	})
```
volmedo, line 205: nit: it may be interesting to add a test for the missing case, the one where the blob has been received but is allocated in a different space, for the sake of completeness

pkg/server/server.go
```diff
@@ -17,7 +17,7 @@ import (
 
 var log = logging.Logger("server")
 
-// ListenAndServe creates a new indexing service HTTP server, and starts it up.
+// ListenAndServe creates a new storage node HTTP server, and starts it up.
```
volmedo, line 20: 💯

volmedo (approved):
LGTM!

Left a non-blocking comment and a couple questions that don't need to be answered now. Is this flow documented somewhere I could take a look to learn more?

pkg/service/storage/ucan.go
alanshaw: Yes originally this was for billing purposes. In the case of failed uploads the user may try to upload the same data again, or sometimes they just uplaod the same thing multiple times. However we don't want to change them multiple times so the `size` value accounts for this. We need to revisit what this means in a decentralized network, as this particular node does not have a view over the whole network and there is no guarantee a second allocation will come to the same node as before (would be useful though).

pkg/service/storage/ucan.go
alanshaw: Yes, exactly.

pkg/service/storage/ucan_test.go
alanshaw: Oh yes good shout https://github.com/storacha/storage/pull/20/commits/ab8c03c91712adcc28bebca0d917bbe8aa03749a

pkg/service/storage/ucan.go
volmedo: thanks for the context. I thought we were looking for previous allocations of the same blob in other spaces too, but I see now the check is scoped only to the space the invocation refers to.

about the issue in the decentralized scenario, would it be a possibility to store the DIDs of the spaces where a CID lives as part of the metadata when indexing? Storage nodes could then check with the indexing-service whether a given blob has already been allocated in that same space. I'm still missing a lot of context, so please let me know if that's just nonsensical 😄.

alanshaw: Yes sounds like a reasonable suggestion!

alanshaw: I've opened an issue with more details. https://github.com/storacha/storage/issues/23 - would you mind adding your thoughts?

PR #22, opened 2024-11-04
PDP

hannahhoward:
# Goals

Implement PDP through curio

# Implementation

The basic strategy here is as follows:
1. When PDP is present, it replaces a number of parts of the blob service, as they now happen directly through curio. (I actually toyed with mirroring the Access and Presigner APIs but it was a little awkward
1. The big chunk of new logic is the aggregator, which is kind of like a mini-w3filecoin pipeline but a lot simpler. I've built this with the goal of making it runnable on SQS queues just like the w3filecoin pipeline, or runnable locally with the job queue. The aggregator can be viewed in terms of its constituent packages:
   1. `pkg/pdp/aggregator/aggregate` handles calculating aggregate roots and assembling piece proofs (it's like a mini go-data-segment
   1. `pkg/pdp/aggregator/fns` contains the core logic of each aggregation step, minutes side effects/state
   1. `pkg/pdp/aggregator/steps.go` contains the actual steps in aggregation with side effects
   1. `pkg/pdp/aggregator/local.go` contains the implementation of aggregation that can run on a native piece of hardware. AWS version to come later
1. PieceAdder is another function that handles essentially saying to Curio "I have a new piece to upload" and Curio responds with an upload URL -- it's relatively analogous to our blob/allocate actually (except we do it in this node)
1. PieceFinder gets the piece cid from the blob hash and size. Importantly, this is probably the ugliest bit of logic for now cause Curio isn't currently setup for reliable read on write, so there is a change we will get blob/accept BEFORE we Curio is ready to give us the piece cid -- that's why there's the retry loop in there for now
1. `pkg/pdp/curio` is just the curio client ot make http calls to curio
1. I also implemented the ReceiptStore cause I needed it
1. I also added a utility for storing ipld types in a data store -- see `pkg/internal/ipldstore` -- I think this will be useful in future
1. There are essentially one new invocation available on the UCAN endpoint - `pdp/info` which you can use to get the state of your aggregation and the inclusion proof. This may still need some changes on auth though? Not sure
1. `pdp/accept` is a self issued invocation used by the server to track when PDP is done and save results.

The PR depends on the following:
https://github.com/storacha/go-ucanto/pull/31
https://github.com/storacha/go-capabilities/pull/2
https://github.com/storacha/go-capabilities/pull/3
as well as two new repos:
 https://github.com/storacha/go-piece
https://github.com/storacha/go-jobqueue (extracted from indexing-service)


cmd/main.go
```diff
@@ -149,6 +168,45 @@ func main() {
...
+							return fmt.Errorf("parsing curio URL: %w", err)
+						}
+						if cCtx.IsSet("curio-auth") {
+							return errors.New("curio-auth must be set if curio is used")
+						}
+						if cCtx.IsSet("pdp-proofset") {
+							return errors.New("pdp-proofset must be set if curio is used")
+						}
+						curioAuth := cCtx.String("curio-auth")
+						proofSet := cCtx.Int64("pdp-proofset")
```
alanshaw, line 194: Use `cCtx.Uint64`?

pkg/pdp/aggregator/aggregate/aggregate.ipldsch
```diff
@@ -0,0 +1,17 @@
...
+
+type AggregatePiece struct {
+	Link           PieceLink
+	InclusionProof ProofData
+}
+
+type Aggregate struct {
+	Root   PieceLink
+	Pieces [AggregatePiece]
+}
```
alanshaw, line 17: For consistency elsewhere:

```suggestion
type AggregatePiece struct {
	link           PieceLink
	inclusionProof ProofData
}

type Aggregate struct {
	root   PieceLink
	pieces [AggregatePiece]
}
```

pkg/pdp/aggregator/fns/fns.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, fmt.Errorf("generating invocation: %w", err)
+		}
+		ok := result.Ok[pdp.PDPAcceptOk, ipld.Builder](pdp.PDPAcceptOk{
+			Aggregate:      aggregate.Root,
+			InclusionProof: aggregatePiece.InclusionProof,
+			Piece:          aggregatePiece.Link,
+		})
+		rcpt, err := receipt.Issue(issuer, ok, ran.FromInvocation(inv))
+		if err != nil {
+			return nil, fmt.Errorf("issueing receipt: %w", err)
```
alanshaw, line 126:
```suggestion
			return nil, fmt.Errorf("issuing receipt: %w", err)
```

pkg/pdp/aggregator/fns/piecessofar.ipldsch
```diff
@@ -0,0 +1,6 @@
+type PieceLink link
+
+type PiecesSoFar struct {
+	TotalSize           uint64
+	ReverseSortedPieces []PieceLink
```
alanshaw, line 5:
```suggestion
	totalSize           Int
	reverseSortedPieces []PieceLink
```

pkg/pdp/aggregator/fns/fns.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	piecesSoFarTS = ts
+}
+
+func PiecesSoFarType() schema.Type {
+	return piecesSoFarTS.TypeByName("PiecesSoFarType")
+}
+
+// PiecesSoFar tracks in progress work building an aggregation
+type PiecesSoFar struct {
```
alanshaw, line 44: We call this a "buffer" in current filecoin pipeline...IDK maybe worth naming it something along those lines?

alanshaw (commented):
Got through about half - have left some minor feedback. Need to 💤

pkg/pdp/aggregator/fns/fns.go
hannahhoward: like that

pkg/pdp/aggregator/fns/fns.go
```diff
@@ -0,0 +1,147 @@
...
+		return buffer, &aggregate, err
+	}
+
+	newSize := buffer.TotalSize + newPiece.PaddedSize()
+	newPieces := InsertReverseSortedBySize(buffer.ReverseSortedPieces, newPiece)
+
+	// if we have reached the minimum aggregate size, submit and start over
+	if newSize >= MinAggregateSize {
+		aggregate, err := aggregate.NewAggregate(newPieces)
+		return Buffer{}, &aggregate, err
```
alanshaw, line 78: If there was an error here, shouldn't we return the buffer untouched? Otherwise we could lose pieces.

```suggestion
		if err != nil {
			return buffer, nil, err
		}
		return Buffer{}, &aggregate, err
```

pkg/service/storage/ucan.go
```diff
@@ -113,28 +132,60 @@ func NewUCANServer(storageService Service) (server.ServerView, error) {
...
+						// get a download url
+						loc = storageService.PDP().PieceFinder().URLForPiece(piece)
+						// submit the piece for aggregation
+						err = storageService.PDP().Aggregator().AggregatePiece(ctx, piece)
+						if err != nil {
+							log.Errorf("submitting piece for aggregation: %w", err)
+							return blob.AcceptOk{}, nil, failure.FromError(err)
+						}
+						// generate the invocation that will complete when aggregation is complete and the piece is accepted
+						pieceAccept, err := pdp.PDPAccept.Invoke(
```
alanshaw, line 169: Musing on this - what if we expose a more generic `blob/retain` that can be used for both? PDP is not really an operation. What I think we really need is an instruction for the blob to continue to be stored over time...and the receipt for an invocation to do this would be details of how to prove it is continually being stored. i.e. a retention receipt would specify the proof type is PDP and contains the inclusion proof. The retention reciept for non-PDP is just "trust me I have it" or "test me with station".

The dictionary definition of "retain" is "continue to have (something); keep possession of".

Side note: perhaps the invocation could include accepted retention policies?

After a blob has been accepted (delivered), the upload service would issue a `blob/retain` invocation to ensure the blob continues to persist on the storage node. Of course for PDP the invocation may have some async tasks to perform (the aggregation), so perhaps simply responds with a `blob/retain/start` effect (not sold on that name).

We can surface this to users by adding the `blob/retain/start` fx to the original `blob/add` invocation.

"retain" also has a nice antonym "release", which we might use to stop the node from submitting PDP proofs (and allow it to garbage collect).

alanshaw (approved):
I don't think I have any blocking feedback here. I think it's fine for now, but the current way it is integrated could use a clean up.

I'm interested in your thoughts on the `blob/retain` idea (see comments) as it feels like a part we're missing in the flow and might allow PDP to land here quite nicely IMHO.

pkg/pdp/aggregator/fns/fns.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Buffer tracks in progress work building an aggregation
+type Buffer struct {
+	TotalSize           uint64
+	ReverseSortedPieces []piece.PieceLink
+}
+
+// InsertReverseSortedBySize adds a piece to a list of pieces sorted largest to smallest, maintaining sort order
+func InsertReverseSortedBySize(sortedPieces []piece.PieceLink, newPiece piece.PieceLink) []piece.PieceLink {
```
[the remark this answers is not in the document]
hannahhoward: makes sense

pkg/pdp/aggregator/fns/fns.go
```diff
@@ -0,0 +1,147 @@
...
+	pos, _ := slices.BinarySearchFunc(sortedPieces, newPiece, func(test, target piece.PieceLink) int {
+		// flip ordering comparing size cause we're going in reverse order
+		return cmp.Compare(target.PaddedSize(), test.PaddedSize())
+	})
+	return slices.Insert(sortedPieces, pos, newPiece)
+}
+
+// MinAggregateSize is 128MB
+// Max size is 256MB -- this means we will never see an individual piece larger
```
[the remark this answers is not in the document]
hannahhoward: this is a question for @magik6k and a curio thing. we can sort it later.

pkg/pdp/aggregator/steps.go
```diff
@@ -0,0 +1,192 @@
...
+func (pa *PieceAccepter) AcceptPieces(ctx context.Context, aggregateLinks []datamodel.Link) error {
+	aggregates := make([]aggregate.Aggregate, 0, len(aggregateLinks))
+	for _, aggregateLink := range aggregateLinks {
+		aggregate, err := pa.aggregateStore.Get(ctx, aggregateLink)
+		if err != nil {
+			return fmt.Errorf("reading aggregates: %w", err)
+		}
+		aggregates = append(aggregates, aggregate)
+	}
+	// TODO: Should we actually send a piece accept invocation? It seems unneccesary it's all the same machine
```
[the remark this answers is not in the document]
hannahhoward: ah something to consider for future. receipts end up stored already

pkg/pdp/aggregator/steps.go
```diff
@@ -0,0 +1,192 @@
...
+	receipts, err := fns.GenerateReceiptsForAggregates(pa.issuer, aggregates)
+	if err != nil {
+		return fmt.Errorf("generating receipts: %w", err)
+	}
+	for _, receipt := range receipts {
+		if err := pa.receiptStore.Put(ctx, receipt); err != nil {
```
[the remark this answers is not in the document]
hannahhoward: 256 mb should at least be less :)

pkg/service/storage/ucan.go
```diff
@@ -70,10 +76,23 @@ func NewUCANServer(storageService Service) (server.ServerView, error) {
...
+							log.Errorf("signing upload URL: %w", err)
+							return blob.AllocateOk{}, nil, failure.FromError(err)
+						}
+					} else {
+						// use pdp service upload
+						url, err = storageService.PDP().PieceAdder().AddPiece(ctx, digest, cap.Nb().Blob.Size)
+						if err != nil {
+							log.Errorf("adding to pdp service: %w", err)
+							return blob.AllocateOk{}, nil, failure.FromError(err)
+						}
```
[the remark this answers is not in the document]
hannahhoward: welcome to golang -- we could use our Wrap and Unwrap on our result type... but I don't know if it ends up being worth it.

pkg/service/storage/ucan.go
hannahhoward: to be clear, pdp/accept here is is a LATER thing -- it's saying, this will be done when I've stored successfully in PDP.

Happy to iterate on naming -- I don't love all this as it is.

pkg/pdp/aggregator/steps.go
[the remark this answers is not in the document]
alanshaw: Oh, yes of course...way smaller aggregates.

PR #34, opened 2025-02-12
feat: error reporting to Sentry

### LINKED ISSUE #29: Instrument with error reporting
- add whatever library is needed to send errors to Sentry
- make sure it gets called at appropriates (return err from service)
- setup appropriate env vars and configs for staging and prod

volmedo:
Resolves #29 

Set up everything so that our lambdas report errors to Sentry.

### DIFF
--- a/deploy/.env.tpl
+++ b/deploy/.env.tpl
@@ -29,7 +29,7 @@ TF_VAR_region=us-west-2
 # Tags applied to AWS resources (useful for cost accounting).
 TF_VAR_app=storage
 TF_VAR_owner=storacha
-TF_VAR_team=Storacha Engineer
+TF_VAR_team=Storacha Engineering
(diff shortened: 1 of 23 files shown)


alanshaw (approved):
LGTM

deploy/.env.tpl
```diff
@@ -29,7 +29,7 @@ TF_VAR_region=us-west-2
 # Tags applied to AWS resources (useful for cost accounting).
 TF_VAR_app=storage
 TF_VAR_owner=storacha
-TF_VAR_team=Storacha Engineer
+TF_VAR_team=Storacha Engineering
```
alanshaw, line 32: Change default in `variables.tf` also?

volmedo: good point 💯

PR #35, opened 2025-02-12
test: noop blobstore

### LINKED ISSUE #33: AWS Billing Test: fake blob store
# What

Currently, blob/accept contains a check to make sure a blob is present in the blob store before it is accepted.

We should make a version of the blob store (https://github.com/storacha/storage/blob/main/pkg/store/blobstore/interface.go) whose get method never returns an error, and whose put method is a no-op. The get should return nil for actual data

# Why

We want to run an aws billing test that simulates all of the infra costs for the services we will run ourselves in a warm storage tier. However, the storage node is not run by us, and moreover, we can simulate the upload pipeline without actually doing data transfer to the storage node, which will enable us to test MUCH faster.

This will be the code required to make the storage node behave as if it received the data without actually receiving the data -- note in a real world scenario they'd need to submit PDP proofs so this isn't an actual attack vector.

This code probably shouldn't get merged to main.

volmedo:
Resolves #33 

This PR adds a new command line flag `--use-noop-blobstore` that can be used to start the storage node with the blobstore disabled. This is useful in testing scenarios to speed tests up and avoid storage and data transfer charges.

Alternatively, setting the env variable `STORAGE_USE_NOOP_BLOBSTORE=1` will have the same effect.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -59,6 +59,12 @@ var StartCmd = &cli.Command{
 			Usage:   "A delegation that allows the node to cache claims with the indexing service.",
 			EnvVars: []string{"STORAGE_INDEXING_SERVICE_PROOF"},
 		},
+		&cli.BoolFlag{
+			Name:    "use-noop-blobstore",
+			Value:   false,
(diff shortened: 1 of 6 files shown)


frrist:
closing as this is stale, feel free to re-open if required.

PR #43, opened 2025-02-21
test: add unit testing for pdp package

frrist:
## What
- closes https://github.com/storacha/storage/issues/39

## How to Review
Each commit is reviewable on it's own.


pkg/pdp/aggregator/fns/fns_test.go
```diff
@@ -0,0 +1,205 @@
...
+		},
+		{
+			name: "Single piece >256MB (if code permits) => immediate flush or error",
+			// TODO(forrest): do we expect this to be an error case?
+			// The aggregator code comment suggests >256MB is out-of-scope, but not
+			// strictly enforced. Some implementations could treat this
+			// as an error or just flush.
+			pieceSizes:         []int64{300 * MB},
+			expectedBufferSize: 0,
+			expectedAggCount:   1,
```
frrist, line 165: Unsure if this case is realistic, `AggregatePieces` states a pieces should never be larger than 256MB, but doesn't enforce the constraint.

pkg/pdp/aggregator/steps.go
```diff
@@ -55,38 +55,52 @@ func NewInProgressWorkspace(store store.Store) InProgressWorkspace {
...
 type QueueSubmissionFn func(ctx context.Context, aggregateLink datamodel.Link) error
 
+type PieceAggregatorOption func(pa *PieceAggregator)
+
+func WithAggregator(a BufferedAggregator) PieceAggregatorOption {
+	return func(pa *PieceAggregator) {
+		pa.aggregator = a
+	}
+}
+
```
frrist, line 65: Allows the aggregator implementation to be subbed out for a mock in testing.

pkg/pdp/aggregator/pieceaggregator_test.go
```diff
@@ -0,0 +1,166 @@
+package aggregator_test
```
frrist, line 1: Seeking opinions on testing strategy here. So much of this is mocked that all we are really doing is asserting the right methods are called in the right order. While this add some value, my feeling is an integration/system test would be more useful here, ideally ad a higher level in the call stack.

pkg/pdp/curio/client.go
```diff
@@ -15,6 +15,20 @@ import (
 	"github.com/storacha/go-ucanto/principal"
 )
 
+type PDPClient interface {
```
frrist, line 18: Defined interface for Curio client, allowing it to be mocked in `curiofinder_test` and `curioadder_test`.

pkg/pdp/aggregator/pieceaggregator_test.go
alanshaw: @frrist there is https://github.com/storacha/testthenetwork - which runs real storage node, indexer and IPNI, mocking out only the upload service. Would be great to also have PDP storage node tested here.

alanshaw (approved):

PR #51, opened 2025-03-15
feat: implement persisted job queue

### LINKED ISSUE #49: [SPIKE] Persisted Job Queue or Retry Logic
## Problem Statement:
When aggregating pieces or submitting aggregations, failures may occur due to software-related issues or unexpected storage node restarts.

## Expected Behavior:
Jobs in the "job queue" must be retried until they succeed. If a job fails, it should either:
- Be retrieved from its respective store and reprocessed, or
- Persist in the queue (e.g., saved to disk) to ensure it remains available for retrying.

At present when jobs fail, or the node restarts the contents of the queue is lost.

## Ideas
One approach may be to implement or adopt a job queuing library with persistence. (e.g. https://github.com/maragudk/goqite)
Another option could be to persist jobs to a store (we already have one) in a pending state, and ensure they are completed across restarts and failures.

frrist:
- closes #49
- thank you OS contributor: https://github.com/maragudk/goqite/
  - Licence and credit at top of respective files
  - modified from source to allow generics


cmd/start.go
```diff
@@ -156,10 +157,16 @@ var StartCmd = &cli.Command{
 			if err != nil {
 				return err
 			}
+			// TODO(forrest): eventually add option to use a file via jobqueue.NewDB()
+			pdpDB, err := jobqueue.NewInMemoryDB(cCtx.Context)
+			if err != nil {
+				return err
+			}
```
frrist, line 164: I've stubbed this in for now, meaning the job queue isn't yet "persisted", rather it runs a sqlite database in-memory.

cmd/client.go
```diff
@@ -65,6 +60,10 @@ var ClientCmd = &cli.Command{
 				if err != nil {
 					return fmt.Errorf("reading blob file: %w", err)
 				}
+				spaceDid, err := did.Parse(cCtx.String("space-did"))
```
hannahhoward, line 63: reason for move? just curious

pkg/pdp/aggregator/fns/fns.go
```diff
@@ -104,6 +114,14 @@ func AggregatePieces(buffer Buffer, pieces []piece.PieceLink) (Buffer, []aggrega
 }
 
 func SubmitAggregates(ctx context.Context, client *curio.Client, proofSet uint64, aggregates []aggregate.Aggregate) error {
+	log.Info("submit aggregates",
```
hannahhoward, line 117: thank you in general for more logs

pkg/pdp/aggregator/jobqueue/internal/testing/testing.go
```diff
@@ -0,0 +1,73 @@
...
+
+package testing
+
+import (
+	"database/sql"
+	_ "embed"
+	"fmt"
+	"testing"
+
+	_ "github.com/ncruces/go-sqlite3"
```
hannahhoward, line 16: curious how you picked this sqlite driver. there are so many. mostly curious cause I need to pick one for the upload client

pkg/pdp/aggregator/jobqueue/queue/queue.go
```diff
@@ -0,0 +1,243 @@
...
+
+// Send a Message to the queue with an optional delay.
+func (q *Queue) Send(ctx context.Context, m Message) error {
+	return internalsql.InTx(q.db, func(tx *sql.Tx) error {
+		return q.SendTx(ctx, tx, m)
+	})
+}
+
+// SendTx is like Send, but within an existing transaction.
+func (q *Queue) SendTx(ctx context.Context, tx *sql.Tx, m Message) error {
```
hannahhoward, line 106: do we use all the queue methods? if no should we maintain from where this was copied?

pkg/pdp/aggregator/jobqueue/serializer/serializer.go
```diff
@@ -0,0 +1,29 @@
+package serializer
+
+import (
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+	"github.com/ipld/go-ipld-prime/schema"
+	"github.com/storacha/go-ucanto/core/ipld/codec/cbor"
+)
+
+type Serializer[T any] interface {
```
hannahhoward, line 9: It would be nice if we didn't keep writing IPLD serialize / deserialize mechanism.

we have the Builder interface, the schema parsers, etc. Not your fault, let's just keep it moving

pkg/pdp/aggregator/jobqueue/worker/worker.go
```diff
@@ -0,0 +1,261 @@
+// Copyright (c) https://github.com/maragudk/goqite
```
hannahhoward, line 1: this is a lot of code -- is there a danger of maintainence burder to those not familia with the code base it's copied from?

hannahhoward (approved):
minor concern amount about amount of code added just to get a persisted queue -- is there a reason we can't use the original package that is copied?

ultimately no hard block cahnges, LGTM

pkg/pdp/aggregator/jobqueue/internal/testing/testing.go
frrist: Primary motivation was it's pure go and thus cgo-free - really would like to avoid cgo in this repo. 

I reviewed the options outlined here: https://github.com/cvilsmeier/go-sqlite-bench and this package was one of the best performers in the pure-go category.

pkg/service/storage/ucan.go
```diff
@@ -242,11 +242,14 @@ func NewUCANServer(storageService Service, options ...server.Option) (server.Ser
...
-						return blob.AcceptOk{}, nil, failure.FromError(err)
-					}
+					/*
+						err = storageService.Claims().Publisher().Publish(ctx, claim)
+						if err != nil {
+							log.Errorf("publishing location commitment: %w", err)
+							return blob.AcceptOk{}, nil, failure.FromError(err)
+						}
+
+					*/
```
frrist, line 252: Note to self, need to uncomment this before merging.

pkg/pdp/aggregator/jobqueue/queue/queue.go
frrist: There are ~2 methods on `Queue` we don't call here: `Send` and `SendAndGetID`. Also, there are certainly some method that are only called by `Queue` internally that we could unexport. Happy to modify this to make the overall port smaller, and exported methods less. 

> should we maintain from where this was copied

I lean towards no for a couple reasons:
1. I've modified the initial port to include details on how many times a message has been received - allows the worker to provide better viability on task execution.
2. I expect we may move away from sqlite in the future, at which point we'll need to port this anyways. 
3. More generally, I have a hunch further modifications may be needed here (within the jobqueue package, not 'here' specifically) in the future.

pkg/pdp/aggregator/jobqueue/worker/worker.go
frrist: This package has evolved from the initial port. Primarily via the introduction of generics+serializer and a logging interface that aligns with our own. I chose to leverage the original project’s functionality and tests, then adapt it for our needs without reinventing the wheel/writing from scratch. The goal here was saving us some development time.

cmd/client.go
frrist: Purely accidental (was basing this on #52), will revert.

pkg/pdp/aggregator/jobqueue/serializer/serializer.go
frrist: Is there a cleaner alternative here? Perhaps this sort of thing should live in go-libstoracha?

frrist:
> minor concern amount about amount of code added just to get a persisted queue -- is there a reason we can't use the original package that is copied?

Hopefully provided enough context in my replies to motivated the porting, primary motivation was getting something I had confidence in working, then modifying to meet our needs while saving development time.

Aside: I will be sure to address the breakages in the lambda code, will also look at updating the make file to catch that locally when building.

PR #52, opened 2025-03-18
refactor: migrate remaining cli commands to use key-file

frrist:
a follow on to https://github.com/storacha/storage/pull/46 that moves rest of commands to --key-file pattern.


hannahhoward (approved):
LGTM, though we should speak to the team about other repos, and decide on a consistent approach.

PR #57, opened 2025-04-04
feat: implement replicator for replication protocol

frrist:
- Follows https://github.com/storacha/go-libstoracha/pull/14 and https://github.com/storacha/storage/pull/56

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -21,13 +21,14 @@ import (
 	"github.com/storacha/go-ucanto/principal"
 	ed25519 "github.com/storacha/go-ucanto/principal/ed25519/signer"
 	ucanserver "github.com/storacha/go-ucanto/server"
+	"github.com/urfave/cli/v2"
+
 	"github.com/storacha/storage/cmd/enum"
 	"github.com/storacha/storage/pkg/presets"
 	"github.com/storacha/storage/pkg/principalresolver"
 	"github.com/storacha/storage/pkg/server"
 	"github.com/storacha/storage/pkg/service/storage"
 	"github.com/storacha/storage/pkg/store/blobstore"
-	"github.com/urfave/cli/v2"
 )
 
 var StartCmd = &cli.Command{
@@ -201,6 +202,24 @@ var StartCmd = &cli.Command{
 			indexingServiceURL = *u
 		}
 
+		uploadServiceDID := presets.UploadServiceDID
+		if os.Getenv("STORAGE_UPLOAD_SERVICE_DID") != "" {
+			d, err := did.Parse(os.Getenv("STORAGE_UPLOAD_SERVICE_DID"))
+			if err != nil {
+				return fmt.Errorf("parsing indexing service DID: %w", err)
+			}
+			uploadServiceDID = d
+		}
+
+		uploadServiceURL := *presets.UploadServiceURL
+		if os.Getenv("STORAGE_UPLOAD_SERVICE_URL") != "" {
+			u, err := url.Parse(os.Getenv("STORAGE_UPLOAD_SERVICE_URL"))
+			if err != nil {
+				return fmt.Errorf("parsing indexing service URL: %w", err)
+			}
+			uploadServiceURL = *u
+		}
+
 		var indexingServiceProofs delegation.Proofs
 		if cCtx.String("indexing-service-proof") != "" {
 			dlg, err := delegation.Parse(cCtx.String("indexing-service-proof"))
@@ -218,6 +237,7 @@ var StartCmd = &cli.Command{
 			storage.WithPublisherDatastore(publisherDs),
 			storage.WithPublicURL(*pubURL),
 			storage.WithPublisherDirectAnnounce(announceURL),
+			storage.WithUploadServiceConfig(uploadServiceDID, uploadServiceURL),
 			storage.WithPublisherIndexingServiceConfig(indexingServiceDID, indexingServiceURL),
 			storage.WithPublisherIndexingServiceProof(indexingServiceProofs...),
 			storage.WithReceiptDatastore(receiptDs),
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.14.0
 	github.com/multiformats/go-multibase v0.2.0
 	github.com/multiformats/go-multihash v0.2.3
-	github.com/storacha/go-libstoracha v0.0.1
+	github.com/storacha/go-libstoracha v0.0.2
 	github.com/storacha/go-ucanto v0.3.0
 	github.com/stretchr/testify v1.10.0
 	github.com/urfave/cli/v2 v2.27.5
--- a/pkg/presets/presets.go
+++ b/pkg/presets/presets.go
@@ -10,6 +10,8 @@ var (
 	AnnounceURL, _        = url.Parse("https://cid.contact/announce")
 	IndexingServiceDID, _ = did.Parse("did:web:indexer.storacha.network")
 	IndexingServiceURL, _ = url.Parse("https://indexer.storacha.network")
+	UploadServiceURL, _   = url.Parse("https://up.storacha.network")
+	UploadServiceDID, _   = did.Parse("did:web:up.storacha.network")
 	PrincipalMapping      = map[string]string{
 		"did:web:staging.up.storacha.network": "did:key:z6MkhcbEpJpEvNVDd3n5RurquVdqs5dPU16JDU5VZTDtFgnn",
 		"did:web:up.storacha.network":         "did:key:z6MkqdncRZ1wj8zxCTDUQ8CRT8NQWd63T7mZRvZUX8B7XDFi",
--- a/pkg/service/replicator/replicate.go
+++ b/pkg/service/replicator/replicate.go
@@ -1,0 +1,85 @@
+package replicator
+
+import (
+	"context"
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/storacha/go-libstoracha/jobqueue"
+	"github.com/storacha/go-ucanto/client"
+	"github.com/storacha/go-ucanto/principal"
+
+	"github.com/storacha/storage/pkg/pdp"
+	"github.com/storacha/storage/pkg/service/blobs"
+	"github.com/storacha/storage/pkg/service/claims"
+	replicahandler "github.com/storacha/storage/pkg/service/storage/handlers/replica"
+	"github.com/storacha/storage/pkg/store/receiptstore"
+)
+
+var log = logging.Logger("replicator")
+
+type Replicator interface {
+	Replicate(context.Context, *replicahandler.TransferRequest) error
+}
+
+type Service struct {
+	queue *jobqueue.JobQueue[*replicahandler.TransferRequest]
+}
+
+type adapter struct {
+	id         principal.Signer
+	pdp        pdp.PDP
+	blobs      blobs.Blobs
+	claims     claims.Claims
+	receipts   receiptstore.ReceiptStore
+	uploadConn client.Connection
+}
+
+func (a adapter) ID() principal.Signer                { return a.id }
+func (a adapter) PDP() pdp.PDP                        { return a.pdp }
+func (a adapter) Blobs() blobs.Blobs                  { return a.blobs }
+func (a adapter) Claims() claims.Claims               { return a.claims }
+func (a adapter) Receipts() receiptstore.ReceiptStore { return a.receipts }
+func (a adapter) UploadConnection() client.Connection { return a.uploadConn }
+
+func New(
+	id principal.Signer,
+	p pdp.PDP,
+	b blobs.Blobs,
+	c claims.Claims,
+	rstore receiptstore.ReceiptStore,
+	uploadConn client.Connection,
+) (*Service, error) {
+
+	replicationQueue := jobqueue.NewJobQueue[*replicahandler.TransferRequest](
+		jobqueue.JobHandler(func(ctx context.Context, request *replicahandler.TransferRequest) error {
+			return replicahandler.Transfer(ctx,
+				&adapter{
+					id:         id,
+					pdp:        p,
+					blobs:      b,
+					claims:     c,
+					receipts:   rstore,
+					uploadConn: uploadConn,
+				},
+				request)
+		}),
+		jobqueue.WithErrorHandler(func(err error) {
+			log.Errorf("error while handling replication request: %s", err)
+		}),
+	)
+
+	return &Service{queue: replicationQueue}, nil
+}
+
+func (r *Service) Replicate(ctx context.Context, task *replicahandler.TransferRequest) error {
+	return r.queue.Queue(ctx, task)
+}
+
+func (r *Service) Start() error {
+	r.queue.Startup()
+	return nil
+}
+
+func (r *Service) Stop(ctx context.Context) error {
+	return r.queue.Shutdown(ctx)
+}
--- a/pkg/service/storage/options.go
+++ b/pkg/service/storage/options.go
@@ -12,6 +12,7 @@ import (
 	"github.com/storacha/go-ucanto/principal"
 	"github.com/storacha/go-ucanto/transport/http"
 	"github.com/storacha/go-ucanto/ucan"
+
 	"github.com/storacha/storage/pkg/access"
 	"github.com/storacha/storage/pkg/pdp"
 	"github.com/storacha/storage/pkg/presigner"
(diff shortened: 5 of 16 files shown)


pkg/pdp/replicator/replicate.go
```diff
@@ -0,0 +1,231 @@
...
+	})
+	rcpt, err := receipt.Issue(r.storageService.ID(), ok, ran.FromInvocation(task.Invocation))
+	if err != nil {
+		return fmt.Errorf("issuing receipt: %w", err)
+	}
+	if err := r.storageService.Receipts().Put(ctx, rcpt); err != nil {
+		return fmt.Errorf("failed to put transfer receipt: %w", err)
+	}
+	// TODO how does one send a receipt to the indexing service
+	client.Execute(rcpt, r.uploadServiceConn)
```
frrist, line 224: @hannahhoward is there any prior art I can borrow from here? Looking to send a receipt for the transfer invocation to the upload service (sorry my TODO is wrong here)

pkg/pdp/replicator/replicate.go
```diff
@@ -0,0 +1,231 @@
...
+			UcanAwait: blob.Await{
+				Selector: ".out.ok",
+				// TODO IDK what this does, or what to put here, or if we even need it?
+				Link: task.Invocation.Link(),
+			},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to accept replication source blob %s: %w", task.Blob.Digest, err)
+	}
```
frrist, line 187: Taking a dependency on the `BlobAccept` method from the ucan package isn't going to work here. I am attempting to avoid re-writing the same code, so may move some of these methods to a different package that can be imported by both the ucan server, and things like the replicator as I image this won't be the only time we need to re-use this logic.

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/multiformats/go-multiaddr v0.14.0
 	github.com/multiformats/go-multibase v0.2.0
 	github.com/multiformats/go-multihash v0.2.3
-	github.com/storacha/go-libstoracha v0.0.1
+	github.com/storacha/go-libstoracha v0.0.2-0.20250408211255-f4eb984a91fe
```
frrist, line 30: https://github.com/storacha/go-libstoracha/pull/14 needs to land before this, then we I need to update this ref before merge.

pkg/presets/presets.go
```diff
@@ -10,6 +10,8 @@ var (
 	AnnounceURL, _        = url.Parse("https://cid.contact/announce")
 	IndexingServiceDID, _ = did.Parse("did:web:indexer.storacha.network")
 	IndexingServiceURL, _ = url.Parse("https://indexer.storacha.network")
+	UploadServiceURL, _   = url.Parse("https://up.storacha.network")
+	UploadServiceDID, _   = did.Parse("did:web:up.storacha.network")
```
frrist, line 14: To Reviewer: please validate I have defined these correctly, based on code I've seen in other repos.

pkg/service/capabilities/service.go
```diff
@@ -0,0 +1,240 @@
+package capabilities
```
frrist, line 1: Moved these here so the replicator can depend on these methods without a circular dep (it can't depend on the storage service)

pkg/service/replicator/replicate.go
```diff
@@ -0,0 +1,218 @@
...
+	if acceptResp.Piece != nil {
+		// generate the invocation that will complete when aggregation is complete and the piece is accepted
+		pieceAccept, err := cap_pdp.Accept.Invoke(
+			r.id,
+			// TODO validate this is the correct audience
+			r.uploadService.ID().DID(),
+			r.id.DID().String(),
+			cap_pdp.AcceptCaveats{
+				Piece: *acceptResp.Piece,
+			}, delegation.WithNoExpiration())
```
frrist, line 168: to reviewer: need to validate I have the issuer, audience, and resource defined correctly here.

pkg/service/storage/options.go
```diff
@@ -213,6 +215,20 @@ func WithPublisherIndexingServiceConfig(serviceDID ucan.Principal, serviceURL ur
 	}
 }
 
+// WithUploadServiceConfig configures UCAN service invocation details
+// for communicating with the upload service.
+func WithUploadServiceConfig(serviceDID ucan.Principal, serviceURL url.URL) Option {
```
frrist, line 220: required by the replicator.

pkg/service/storage/ucan.go
```diff
@@ -172,6 +179,100 @@ func NewUCANServer(storageService Service, options ...server.Option) (server.Ser
...
+						storageService.ID(),
+						storageService.ID(),
+						storageService.ID().DID().GoString(),
+						replica.TransferCaveats{
+							Space:    cap.Nb().Space,
+							Blob:     cap.Nb().Blob,
+							Location: cap.Nb().Location,
+							Cause:    inv.Link(),
+						},
+					)
```
frrist, line 211: to reviewer: need to validate this is the correct issuer, audience, and resource. Should one or more of these be the upload service?

pkg/service/storage/ucan.go
```diff
@@ -172,6 +179,100 @@ func NewUCANServer(storageService Service, options ...server.Option) (server.Ser
...
+					if err != nil {
+						return replica.AllocateOk{}, nil, failure.FromError(err)
+					}
+
+					if len(lc.Location) < 1 {
+						return replica.AllocateOk{}, nil, failure.FromError(fmt.Errorf("location missing from location claim"))
+					}
+
+					// TODO: which one do we pick if > 1?
+					replicaAddress := lc.Location[0]
```
frrist, line 247: to reviewer: seeking feedback on stronger validation assertions here.

pkg/service/storage/ucan.go
```diff
@@ -172,6 +179,100 @@ func NewUCANServer(storageService Service, options ...server.Option) (server.Ser
...
+						return replica.AllocateOk{}, nil, failure.FromError(err)
+					}
+					for block, err := range inv.Blocks() {
+						if err != nil {
+							return replica.AllocateOk{}, nil, fmt.Errorf("iterating replica allocate invocation blocks: %w", err)
+						}
+						if err := trnsfInv.Attach(block); err != nil {
+							return replica.AllocateOk{}, nil, fmt.Errorf("failed to replica allocate invocation block (%s) to transfer invocation: %w", block.Link().String(), err)
+						}
+					}
```
frrist, line 222: to reviewer: is this the correct way to chain the replica/allocate invocation to the replica/transfer invocation?

pkg/service/storage/ucan.go
```diff
@@ -172,6 +179,100 @@ func NewUCANServer(storageService Service, options ...server.Option) (server.Ser
...
+
+					// TODO: which one do we pick if > 1?
+					replicaAddress := lc.Location[0]
+
+					// FIXME: use a real context, requires changes to server
+					ctx := context.TODO()
+					allocateResp, err := storageService.Capabilities().BlobAllocate(ctx, &capabilities.BlobAllocateRequest{
+						Space: cap.Nb().Space,
+						Blob:  cap.Nb().Blob,
+						Cause: inv.Link(),
```
frrist, line 254: The replica/allocate is the cause of this blob/allocate invocation, yeah? Rather than the transfer?

pkg/service/storage/ucan.go
```diff
@@ -172,6 +179,100 @@ func NewUCANServer(storageService Service, options ...server.Option) (server.Ser
...
+						Space:      cap.Nb().Space,
+						Blob:       cap.Nb().Blob,
+						Source:     replicaAddress,
+						Sink:       allocateResp.Address.URL,
+						Invocation: trnsfInv,
+					}); err != nil {
+						return replica.AllocateOk{}, nil, failure.FromError(fmt.Errorf("failed to enqueue replication task: %w", err))
+					}
+
+					return replica.AllocateOk{Size: allocateResp.Size}, fx.NewEffects(fx.WithFork(fx.FromInvocation(trnsfInv))), nil
```
frrist, line 272: the RFC for this work states:
> The blob/replica/allocate task receipt includes an async task that will be performed by the storage node - blob/replica/transfer. The blob/replica/transfer task is completed when the storage node has transferred the blob from its location to the storage node
>...
> Client can poll the upload service for the blob/replica/transfer receipt.

With the transfer invocation as an fx of replica/allocate will the described polling work?

pkg/service/storage/ucan_test.go
```diff
@@ -5,30 +5,43 @@ import (
...
+
+	"github.com/storacha/storage/pkg/internal/testutil"
 )
 
 func TestServer(t *testing.T) {
 	ctx := context.Background()
 	svc, err := New(WithIdentity(testutil.Alice), WithLogLevel("*", "warn"))
 	require.NoError(t, err)
-	t.Cleanup(func() { svc.Close(ctx) })
+	err = svc.Startup()
```
frrist, line 40: Previously we never stated the storage service, now that it's using a queue for replication, we must start it for the queue to run.

pkg/service/storage/ucan_test.go
```diff
@@ -321,3 +334,213 @@ func TestServer(t *testing.T) {
...
+	agentCh := make(chan message.AgentMessage, 1)
+	mux := http.NewServeMux()
+
+	// Endpoint to serve data.
+	mux.HandleFunc(fmt.Sprintf("/%s", sourcePath), func(w http.ResponseWriter, r *http.Request) {
+		_, _ = w.Write(serveData)
+	})
+	// Endpoint to store data on the replica.
+	mux.HandleFunc(fmt.Sprintf("/%s", sinkPath), func(w http.ResponseWriter, r *http.Request) {
+		require.NoError(t, svc.Blobs().Store().Put(ctx, digest, uint64(len(serveData)), bytes.NewReader(serveData)))
```
frrist, line 503: simulates a the replication upload path.

pkg/service/storage/ucan_test.go
```diff
@@ -321,3 +334,213 @@ func TestServer(t *testing.T) {
...
+	// Endpoint to simulate the upload service.
+	mux.HandleFunc(fmt.Sprintf("/%s", uploadServicePath), func(w http.ResponseWriter, r *http.Request) {
+		roots, blocks, err := car.Decode(r.Body)
+		require.NoError(t, err)
+		bstore, err := blockstore.NewBlockReader(blockstore.WithBlocksIterator(blocks))
+		require.NoError(t, err)
+		agentMessage, err := message.NewMessage(roots, bstore)
+		require.NoError(t, err)
+		agentCh <- agentMessage
+	})
```
frrist, line 515: simulates the upload-service, returning the transfer receipt on the channel

pkg/service/storage/ucan_test.go
```diff
@@ -321,3 +334,213 @@ func TestServer(t *testing.T) {
...
+}
+
+func (f *FakePresigned) SignUploadURL(ctx context.Context, digest multihash.Multihash, size, ttl uint64) (url.URL, http.Header, error) {
+	return f.uploadURL, nil, nil
+}
+
+func (f *FakePresigned) VerifyUploadURL(ctx context.Context, url url.URL, headers http.Header) (url.URL, http.Header, error) {
+	// TODO: implement when needed.
+	panic("implement me")
+}
```
frrist, line 546: injected into the storage service, allowing the upload URL to be configured by the test, pointing at the test HTTP server.

pkg/service/capabilities/service.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// even if a previous allocation was made in this space, we create
+	// another for the new invocation.
+	err = s.blob.Allocations().Put(ctx, allocation.Allocation{
+		Space:   req.Space,
+		Blob:    allocation.Blob(req.Blob),
+		Expires: expiresAt,
+		// REVIEW: is this the correct cause? The invocation link rather than the allocate caveats cause field?
+		Cause: req.Cause,
```
frrist, line 141: similar comments about the `Cause` of this invocation.

cmd/start.go
```diff
@@ -208,6 +209,24 @@ var StartCmd = &cli.Command{
 			indexingServiceURL = *u
 		}
 
+		uploadServiceDID := presets.UploadServiceDID
+		if os.Getenv("STORAGE_UPLOAD_SERVICE_DID") != "" {
+			d, err := did.Parse(os.Getenv("STORAGE_INDEXING_SERVICE_DID"))
```
hannahhoward, line 214:
```suggestion
			d, err := did.Parse(os.Getenv("STORAGE_UPLOAD_SERVICE_DID"))
```

cmd/start.go
```diff
@@ -208,6 +209,24 @@ var StartCmd = &cli.Command{
...
+			if err != nil {
+				return fmt.Errorf("parsing indexing service DID: %w", err)
+			}
+			uploadServiceDID = d
+		}
+
+		uploadServiceURL := *presets.UploadServiceURL
+		if os.Getenv("STORAGE_UPLOAD_SERVICE_URL") != "" {
+			u, err := url.Parse(os.Getenv("STORAGE_INDEXING_SERVICE_URL"))
```
hannahhoward, line 223:
```suggestion
			u, err := url.Parse(os.Getenv("STORAGE_UPLOAD_SERVICE_URL"))
```

pkg/presets/presets.go
hannahhoward: yes, to my knownledge

pkg/service/storage/ucan.go
hannahhoward: yes, the blob/replica/transfer receipt will become available to the client when `ucan/conclude` is invoked with said receipt on the upload service at the end of replication

pkg/service/capabilities/service.go
hannahhoward: I can't decide if these are really a whole "service" -- they really have no internal mutable state.

I almost wonder if you want to instead do this:

- move to something like `pkg/service/storage/handlers/blobs/blobs.go` 
- define the service not as a struct but as an interface with the deps you need (ID()/Blobs()/PDP()/Claims())
- hey the storage service already matches the interface!

- change BlobAllocate / BlobReplicate to pure functions taking the service as the first dependency

- avoids ever having to have a seperate capabilities service (btw, if you do want to keep the service, I would rename it "handlers" or something)

pkg/service/replicator/replicate.go
```diff
@@ -0,0 +1,220 @@
+package replicator
```
hannahhoward, line 1: Similar to above, I would move the private replicate function to `pkg/service/storage/handlers/replicate/replicate.go`
- define an interface for the things it needs on the storage service:
```
ID()
Receipts()
Blobs()
PDP()
Claims()
UploadService()
```
Signature for Replicate becomes Replicate(ctx,service, task)

- Move all the job queue machinery inside the storage service or move it a much smaller service for just the jobqueue

pkg/service/replicator/replicate.go
hannahhoward: no should be:
issuer: the storage node
audience: the storage node
resource: the storage node

see handler in `pkg/service/storage/ucan.go`

However, why is this not itself in prob replicate? why not have acceptResp.Piece be an optional link to an invocation, instead of the Piece itself?

pkg/service/replicator/replicate.go
hannahhoward: mostly focused on seperate of concerns between stateless replication handling and job scheduling

pkg/service/storage/ucan.go
hannahhoward: no I think cause here is the original space/blob/replicate which should be the Cause value in `cap.Nb().Cause`

hannahhoward: cause you are not ACTUALLY invoking blob/allocate here, you're just using a common handler

(I might rename the function `AllocateBlob` rather than `BlobAllocate` to make it more clear it's a command, not a capability)

hannahhoward (changes requested):
In terms of correctness, I think only the pdp/piece/accept is wrong.

However, after merge, I would like to see a follow on ticket with the suggested refactors. Happy to discuss sync.

pkg/service/replicator/replicate.go
frrist: ooo yeah, I like that much better. Done in latest change.

pkg/service/storage/ucan.go
frrist: The AllocationStore called internally by `blobhandler.Allocate` expects a cause link for the `Put` operation. `inv.Link` is the `blob/replica/allocate` invocation, which is technically the cause(?) where `space/blob/replicate` is the cause of `blob/replica/allocate`, right?

hannahhoward (approved):
looks great :)

PR #63, opened 2025-04-23
fix: use correct add root payload struct

frrist:
- required by change in curio: https://github.com/filecoin-project/curio/pull/480


volmedo (approved):
LGTM provided the mock is updated

frrist:
> provided the mock is updated

Ohh great catch thank. I really need to add a make directive for this. Will do so before merging.

PR #64, opened 2025-04-26
Scope PDP Service Testing

frrist:
- In here I have attempted to scope out a testing approach for the PDP Service using mocks and fake chain data


pkg/pdp/service/contract/contract.go
```diff
@@ -0,0 +1,139 @@
+package contract
```
frrist, line 1: A mechanical change that modifies the PDP service to operate on the contract using an interface so we can mock it during testing.

frrist:
@hannahhoward has verbally confirmed that I can merge this and disable the windows CI because no one uses windows.

PR #72, opened 2025-05-02
feat(publisher): support custom blob address

hannahhoward:
# Goals

Properly support scenarios where blobs are available to retrieve at other address all the way to the indexer without requiring additional wiring

# Implementation

- in aws, rather than exposing a redirect endpoint for getblob, we allow the publisher to actually post the correct multiaddr a direct bucket fetch
- for curio, we publish the direct address for resolving curio public piece fetches
- pass options all the way down from storage -> publisher (this process sucks)
- add support for a custom blob address to the publisher


alanshaw (approved):
LGTM

frrist (approved):

PR #73, opened 2025-05-02
feat(config): add configuration system with file support

frrist:
Add a comprehensive configuration system that allows configuring the storage node via:
1. CLI flags (highest precedence)
2. Environment variables
3. TOML configuration file (though json and yaml are also supported)
4. Sensible defaults (lowest precedence)

Includes an example.toml configuration file with commented options.

Motivation: I don't like the UX of needing to pass a mixture of flags and env vars to run. Now there is a flag, env-var, and config file value for everything.


pkg/config/config.go
```diff
@@ -0,0 +1,346 @@
...
+		"upload.service_url": "UPLOAD_SERVICE_URL",
+
+		// PDP
+		"pdp.server_url": "CURIO_URL",
+		"pdp.proof_set":  "PDP_PROOFSET",
+	}
+
+	// Create the aliases for environment variables
+	for key, envVar := range envMappings {
+		if err := v.BindEnv(key, "STORAGE_"+envVar); err != nil {
```
volmedo, line 302: why do we need to use `BindEnv` if we have already used `AutomaticEnv`?

cmd/start.go
```diff
@@ -38,12 +38,23 @@ var StartCmd = &cli.Command{
...
+			Name:      "key-file",
+			Usage:     "Path to a file containing ed25519 private key, typically created by the id gen command.",
+			EnvVars:   []string{"STORAGE_PRIVATE_KEY"},
+			TakesFile: true,
+		},
+		&cli.StringFlag{
+			Name:    "config",
+			Usage:   "Path to configuration file.",
+			EnvVars: []string{"STORAGE_CONFIG"},
+		},
```
volmedo, line 53: should this one be a `cli.PathFlag` too? Also, I don't know what it does, but should this flag also have `TakesFile: true`?

volmedo (commented):
looks good, left some comments

pkg/config/config.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// PDPConfig holds configuration for the Proof of Data Possession subsystem
+type PDPConfig struct {
+	ServerURL string `toml:"server_url" json:"server_url" mapstructure:"server_url"`
+	ProofSet  uint64 `toml:"proof_set" json:"proof_set" mapstructure:"proof_set"`
+}
+
+// Node represents the full configuration for a storage node
+type Node struct {
```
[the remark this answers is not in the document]
frrist: Yeah that's fair, will rename to `NodeConfig`.

pkg/config/config.go
```diff
@@ -0,0 +1,346 @@
...
+		"indexing.service_url":   "INDEXING_SERVICE_URL",
+		"indexing.announce_url":  "ANNOUNCE_URL",
+
+		// Upload
+		"upload.service_did": "UPLOAD_SERVICE_DID",
+		"upload.service_url": "UPLOAD_SERVICE_URL",
+
+		// PDP
+		"pdp.server_url": "CURIO_URL",
+		"pdp.proof_set":  "PDP_PROOFSET",
```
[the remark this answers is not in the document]
frrist: Yeah... I don't love this either. Defining constants seems like a better option for now -- will make that change.

Note: this change introduces `viper` a configuration package. Typically Viper is used together with [`cobra`](https://github.com/spf13/cobra), a CLI package. Given we are currently using ufave/cli everywhere I didn't want to introduce cobra (yet) as it will add to teams cognitive overhead when dev-ing here. Iff we move to cobra in the future, this mapping will go away as cobra+viper have a pattern for it. All this to say: this mapping is somewhat temporary, and can be removed by replacing ufave/cli with cobra in the future.

pkg/config/config.go
frrist: ahh good catch, this binding call isn't necessary -- will remove.

cmd/start.go
frrist: Yeah, probably best to include that in the flags definition. ufave/cli docs are a bit sparse on what this option does, my gut says its related to shell completion: https://github.com/urfave/cli/blob/8f20a3568191ab2a77de5bf37244ad607098c301/flag_impl.go#L55

Will change.

frrist:
Pausing work here until this becomes higher priority.

alanshaw:
🙏 can we please ensure we hide esoteric flags in default help output as we discussed here https://github.com/storacha/storage/pull/78#discussion_r2101267859

PR #76, opened 2025-05-13
Implement terraform to deploy storage node on EC2 instance

frrist:
In draft until fix is made to mounting EBS volume for storage node repo.
Closes https://github.com/storacha/project-tracking/issues/334


deploy/storage/nginx.conf.tpl
```diff
@@ -0,0 +1,43 @@
+server {
```
frrist, line 1: handle TLS termination to the storage node.

PR #77, opened 2025-05-19
fix: handle task engine restarts grecefully

frrist:
# Summary
  - Refactor PDP service architecture with improved database abstraction, better error handling, and enhanced testing infrastructure
  - Fix critical bugs in task engine restarts and proving period calculations that affected system reliability

### Database & Storage Improvements
  - Added database abstraction layer (pkg/database/) with support for both Raw SQLite and SQLite via GORM ORM backends
  - Migrated to SQLite as default with proper foreign key constraint support for better development experience
  - Enhanced data models with improved triggers and better state management
  - Fixed GORM configuration for auto-incrementing primary keys and proper driver imports

### PDP Service Architecture Refactor
  - Redesigned task scheduler engine with better state management and graceful restart handling
  - Improved contract integration with proper mock generation and cleaner interfaces
  - Enhanced middleware for better error handling and request logging

### Bug Fixes & Reliability
  - Fixed task engine restart issues that caused proving period calculation errors
  - Corrected proving period adjustments to prevent scheduling conflicts
  - Improved error handling throughout the PDP pipeline
  - Added proper cleanup for temporary resources and database connections

### Testing Infrastructure
  - Comprehensive test harness (pkg/pdp/testing/) for integration testing
  - Mock generation for all external dependencies (Ethereum clients, contracts, chain clients)
  - Fake chain client for deterministic testing scenarios
  - End-to-end test cases for proof set creation
    - Need to add move coverage for proving an uploading

### Development Experience
  - Added memory-based stores for testing (keystore, blobstore)
  - Enhanced CLI experience with better wallet management and data directory structure
  - Improved build system with comprehensive mock generation
  - Better logging and debugging support throughout the system


pkg/pdp/scheduler/handler.go
```diff
@@ -0,0 +1,265 @@
...
+		shouldCommit, err := extra(tID, tx)
+		if err != nil {
+			return err
+		}
+
+		if shouldCommit {
+			return nil
+		}
+
+		return ErrDoNotCommit
```
alanshaw, line 57: Returning an error that is ignored below made this difficult for me to understand. Was there some functionality that was removed that would have been done when `shouldCommit` is true? Perhaps this needs a comment explaining?

pkg/pdp/scheduler/handler.go
```diff
@@ -0,0 +1,265 @@
...
+			continue
+		}
+		if result.RowsAffected == 0 {
+			// Already taken by someone else (or in race condition). Skip it.
+			log.Debugf("Task %d was already claimed; skipping", id)
+			continue
+		}
+
+		// Successfully claimed this task, so let’s run it in a goroutine:
+		// TODO doing this in parallel is causing concurrency issues with the sqlite database.
```
alanshaw, line 107: YOLO

alanshaw (approved):
❤️ love all the additional logging.

pkg/pdp/scheduler/handler.go
frrist: Actually, good call-out. This has been addressed by:
- the retry logic in `handleDoneTask` and `AddTask`, which will retry db operations that timeout due to locking until success.
- the `doneMu` of `taskTypeHandler` which locks around the db transaction.
- the changes to the `database` package which add `_pragma=journal_mode=wal` and `_pragma=busy_timeout=5000`, using a write ahead log and longer operation timeout. 

SQLite will still get cranky about parallel writes since it uses database-wide locks. 
Of note - previously this engine was designed to use Postgres (yugabyte) - which uses row-level locking. I think with the changes I've mentioned above, we should be good to remove the TODO here (which I'll do before merging)

pkg/pdp/scheduler/handler.go
frrist: Yeah, this is _really_ confusing, I'll add a comment here. To explain:
- `AddTask` attempts to add a task to the database.
- `AddTask` invokes a tasks `Adder` function (`extra` in this method). The `Adder` function, implemented by each tasks, is used to determine if said task should run.
- If the task needs to run, its `Adder` returns `true`, otherwise `false`.
- In the event the `Adder` returns `false` we don't want to run the task, so we return an error here to abort the database transactions which creates the task in the db. 

I'll add a comment to clarify this.

cmd/serve.go
```diff
@@ -77,6 +81,7 @@ var pdpCmd = &cli.Command{
 		},
 	},
 	Action: func(cctx *cli.Context) error {
+		logging.SetLogLevel("*", "INFO")
```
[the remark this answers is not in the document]
frrist: humm, no - not really. Though I do like seeing the info logs as an operator. This only sets the logging level to info when running the node as a PDP node, so I think I'll leave this as is for now.

More generally, we'll need to make some telemetry improvements here (metrics and tracing are missing), will address the inconsistent log levels in a follow on.

PR #78, opened 2025-05-21
feat: configre multiple IPNI announce URLs

alanshaw:
When we deploy our own IPNI node we will want storage nodes to announce to it in addition to cid.contact.

This PR allows multiple IPNI accounce URLs to be configured from the CLI and for AWS deploy.

For the CLI, `STORAGE_ANNOUNCE_URL` is no longer supported, use `STORAGE_IPNI_ANNOUNCE_URLS` instead, which is expected to be a JSON array of URLs.


volmedo (approved):
LGTM

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sqs v1.37.0
 	github.com/aws/aws-sdk-go-v2/service/ssm v1.55.5
 	github.com/awslabs/aws-lambda-go-api-proxy v0.16.2
-	github.com/ethereum/go-ethereum v1.14.13
+	github.com/ethereum/go-ethereum v1.15.11
```
frrist, line 18: humm, can this update be reverted? iirc (I can barely remember) we need to use 1.14.13 specifically for some PDP interactions

frrist (approved):
Please revert the eth dep update if it's not too much trouble. I have some vague memory (which may be wrong) that changing versions introduced issues with out PDP implementation.

cmd/start.go
```diff
@@ -194,13 +194,22 @@ var StartCmd = &cli.Command{
 			return fmt.Errorf("parsing public URL: %w", err)
 		}
 
-		announceURL := *presets.AnnounceURL
-		if os.Getenv("STORAGE_ANNOUNCE_URL") != "" {
-			u, err := url.Parse(os.Getenv("STORAGE_ANNOUNCE_URL"))
+		var ipniAnnounceURLs []url.URL
+		if os.Getenv("STORAGE_IPNI_ANNOUNCE_URLS") != "" {
```
frrist, line 198: nit: Can you also add a flag for this? Personally, I find the inconsistent mixture of flags and env vars for setting configuration to be frustrating as an operator.

alanshaw: To explain that a little - I had previously created _flags_ for configuration options that folks might actually want to change (or are safe to change), and _env vars_ for configurations that are unlikely to need to be configured by your average operator. i.e. the env vars are for power users and testing/debugging/internal use.

In this case I'd not anticipate an average storage node operator to ever need to change these. I can see a case where the defaults change based on the network you're joining - like staging/production, but I'd prefer to expose a '--network' flag that would switch the defaults accordingly.

So, we can expose _every_ configuration value as a flag but I don't think it's the right thing to do - the available options will be overwhelming and we will probably see folks setting configs that are conflict or are incompatible - for example, staging indexer but production upload service.

> I find the inconsistent mixture of flags and env vars for setting configuration to be frustrating as an operator.

I would argue that if there is frustrations we need to figure out what env vars need to be promoted to flags, or what flags need to be created that configure appropriate values, so that we don't allow operators to misconfigure their node and have a bad time.

🙏 If we decide to have everything as a flag then I'd prefer to do this as a separate PR that switches all of them to flags at once.

go.mod
alanshaw: Yes, I will try, I had a bunch of errors though...

cmd/start.go
frrist: Yeah, I understand that approach.

My general preference is a consistent hierarchy: CLI flags > env vars > config file > defaults, with each option available at all levels.

> I would argue that if there is frustrations we need to figure out what env vars need to be promoted to flags, or what flags need to be created that configure appropriate values, so that we don't allow operators to misconfigure their node and have a bad time.

Personally, I'd like a flag for everything. Flags that we don't expect the average user to provide can be marked as `Hidden`, and/or exposed by a help command with verbose output. Libraries like https://github.com/spf13/viper and https://github.com/spf13/cobra make this _trivial_ to implement too.

I'd also _really_ like a config file, and I suppose that's my primary source of frustration, I stated a change for this here even: https://github.com/storacha/storage/pull/73 but have held off for the time being due to priority and library differences.

Regardless, this change is a welcome improvement and my approval is sticky. Happy to tackle this at a later point as you've suggested :+1:

alanshaw: Ok I'm happy with the alternative of have a flag for everything and hide the esoteric ones. Config file also sounds like a good addition 👍 .

PR #82, opened 2025-05-29
chore: use S3 datastore implementation in go-libstoracha

volmedo:
The S3 datastore implementation was moved from here to go-libstoracha. Let's remove it from here and use the one in go-libstoracha from now on.

Replicate capabilities were also updated in go-libstoracha, so some changes were required to adapt the code to the updates.


alanshaw (approved):
LGTM

PR #84, opened 2025-05-30
fix: consistent read for allocations

alanshaw:
I keep seeing this error in the integration tests:

```
service handler {can: "blob/allocate"} error: failed to read allocation after write
```

From here: https://github.com/storacha/storage/blob/2257b678a2bc9755acd6b77a7b2b52e5c4ace5e1/pkg/service/storage/handlers/blob/allocate.go#L149

I think this check is redundant but we don't want reads to _not_ be strongly consistent for allocations.


frrist (approved):
LGTM - though, I think this increase the cost of the dynamoDB operation.

PR #86, opened 2025-06-04
refactor: piri ascii art

alanshaw:
This PR refactors the ASCII art you see when starting the node to read "Piri". It now looks like this:

Screenshot 2025-06-04 at 10 59 06

It's a little less in your face and less tall which I think is better.

As a bonus, this PR also fixes the version number from static text to the version tracked in `version.json`.


Peeja:
OMG, I love this 🌶️

PR #87, opened 2025-06-04
chore: format delegation

frrist:
- returned deletation from piri delegation generat [flags] is formatted as a Base64-encoded CIDv1 with embedded CAR data.
- include replica.AllocateAbility capability in delegation.


hannahhoward (approved):
LGTM

PR #88, opened 2025-06-05
fix: use correct dir for pdp storage dir

frrist:
- previously a typo caused the storage dir to be a temp dir, which really messed up note restarts as a new path and database was used/inited each time.


frrist:
I've merged this w/o review as it's a simple change that I've confirmed fixes the issue, if we decide we want to revert it post merge I am happy to.

Issue #98, opened 2025-06-10
Migrate Replicator to Use SQLite Job Queue System


frrist:
## Description

The replicator service (`pkg/service/replicator/replicate.go`) currently uses an in-memory job queue for handling replication requests. This should be migrated to use the same SQLite-based job queue system used by the aggregator (`pkg/pdp/aggregator/local.go`) for better persistence, reliability, and consistency across the codebase.

## Current Behavior

- Replicator uses `jobqueue.NewJobQueue` from `storacha/go-libstoracha/jobqueue` (in-memory)
- Jobs are lost if the process crashes or restarts
- No persistence of job state or history
- No ability to query job status or retry failed jobs after restart
- Inconsistent with aggregator's approach

## Expected Behavior

1. Use the same SQLite-based job queue system as the aggregator
2. Persist replication jobs across restarts
3. Enable job status tracking and monitoring
4. Support proper retry mechanisms with persistence
5. Maintain consistency across the codebase

### Current Implementation

The replicator (`pkg/service/replicator/replicate.go:52-69`) uses:
```go
replicationQueue := jobqueue.NewJobQueue[*replicahandler.TransferRequest](
    jobqueue.JobHandler(func(ctx context.Context, request *replicahandler.TransferRequest) error {
        return replicahandler.Transfer(ctx, &adapter{...}, request)
    }),
    jobqueue.WithErrorHandler(func(err error) {
        log.Errorf("error while handling replication request: %s", err)
    }),
)
```

### Target Implementation

The aggregator (`pkg/pdp/aggregator/local.go:92-99`) uses:
```go
linkQueue, err := jobqueue.New(
    LinkQueueName,
    db,  // SQLite database
    &serializer.IPLDSerializerCBOR[datamodel.Link]{
        Typ:  &schema.TypeLink{},
        Opts: types.Converters,
    },
    jobqueue.WithLogger(logging.Logger("jobqueue").With("queue", LinkQueueName)),
)
```

### Migration Requirements

1. **Add SQLite Database for Replicator**:
```go
// In pkg/service/replicator/replicate.go
type Service struct {
    queue *jobqueue.JobQueue[*replicahandler.TransferRequest]
    db    database.Database
}

func New(..., dbPath string) (*Service, error) {
    // Create SQLite database
    db, err := sqlitedb.New(dbPath,
        database.WithJournalMode("WAL"),
        database.WithTimeout(5*time.Second),
        database.WithSyncMode(database.SyncModeNORMAL),
    )
[...]

PR #112, opened 2025-06-13
refactor: implement persisted replicator job queue

frrist:
- call back to in-memoiry option when one isn't provided.
- closes https://github.com/storacha/piri/issues/98

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -26,6 +26,8 @@ import (
 	"github.com/urfave/cli/v2"
 
 	"github.com/storacha/piri/cmd/enum"
+	"github.com/storacha/piri/pkg/database"
+	"github.com/storacha/piri/pkg/database/sqlitedb"
 	"github.com/storacha/piri/pkg/presets"
 	"github.com/storacha/piri/pkg/principalresolver"
 	"github.com/storacha/piri/pkg/server"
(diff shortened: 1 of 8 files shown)


cmd/start.go
```diff
@@ -272,6 +274,20 @@ var StartCmd = &cli.Command{
...
+
+		db, err := sqlitedb.New(filepath.Join(replicatorDir, "replicator.db"),
+			database.WithJournalMode("WAL"),
+			database.WithTimeout(5*time.Second),
+			database.WithSyncMode(database.SyncModeNORMAL),
+		)
+		if err != nil {
+			return fmt.Errorf("creating jobqueue database: %w", err)
+		}
+
```
frrist, line 290: I really don't like that this setup code is leaking into our cmd package, but I don't see an alternative with the current design.

pkg/service/storage/service.go
```diff
@@ -258,7 +259,14 @@ func New(opts ...Option) (*StorageService, error) {
...
 
-	repl, err := replicator.New(id, pdpImpl, blobs, claims, receiptStore, uploadServiceConnection)
+	if c.replicatorDB == nil {
+		c.replicatorDB, err = sqlitedb.NewMemory()
+		if err != nil {
+			return nil, fmt.Errorf("creating in-memory replicator db: %w", err)
+		}
+	}
+
+	repl, err := replicator.New(id, pdpImpl, blobs, claims, receiptStore, uploadServiceConnection, c.replicatorDB)
```
frrist, line 269: I'd prefer to make the DB a required configuration for the replicator, but unsure how to make that work with current config design. I could fail here if one isn't provided, but that will break other services that use this constructor, such as `pkg/aws/service.go` which are mostly stateless. On principle, the aws service implementation ought to use some aws specific cloud db, but that's beyond the scope of this implementation.

I am assuming the aws package will be removed at some point, since I think it's mostly for testing anyways? At which point we can re-think how these services are instantiated.

PR #123, opened 2025-06-24
Create CODEOWNERS

frrist:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+@frrist
```
volmedo, line 1: is this the same as this?
```suggestion
*	@frrist
```

frrist: :shrug: github says this current file is valid? So I assume it's fine until more folks wanna join as owners?

volmedo (approved):

.github/CODEOWNERS
alanshaw, line 1:
```suggestion
@frrist
@alanshaw
```

alanshaw (changes requested):

Issue #126, opened 2025-06-25
Root Cause Analysis: PDP Prove Task Failure


frrist:
# Root Cause Analysis: PDP Prove Task Failure

## Issue Summary
The PDP (Proof of Data Possession) Prove Task stops executing after running for a period of time, causing the system to fail to generate and submit proofs for data possession challenges.

## Impact
- Proof generation tasks cease to execute
- System fails to meet proof submission requirements
- Potential penalties or loss of rewards for missing proof deadlines

## Root Cause
The issue stems from a failure in the Ethereum transaction monitoring component (`MessageWatcherEth`) that creates a cascading failure preventing proof tasks from being triggered.

### Failure Chain
1. **Transaction Monitoring Failure**: The `MessageWatcherEth` component experiences network timeouts when fetching transaction receipts from the Ethereum RPC endpoint
2. **Early Exit Behavior**: When a timeout occurs, the watcher's update loop exits entirely instead of continuing to process remaining transactions
3. **Status Update Failure**: Transactions remain in "pending" status indefinitely because the watcher stops processing before updating their status
4. **Proof Task Dependency**: The `ProveTask` only triggers when challenge request transactions are marked as successful (`tx_success = true`)
5. **Task Starvation**: With transactions stuck in pending status, no new prove tasks are created

## Evidence
- **Log Analysis**: Shows PDP Prove tasks stop executing while NextProvingPeriod tasks continue to create new transactions
- **Database State**: 326 pending transactions accumulated in the `message_waits_eth` table
- **Error Pattern**: Repeated "i/o timeout" errors when fetching transaction receipts from lotus API

### Code Location
`pkg/pdp/tasks/watcher_eth.go`
- Lines 119-127: Error handling for transaction receipt fetching

### Current Behavior
```go
receipt, err := mw.api.TransactionReceipt(ctx, txHash)
if err != nil {
    if errors.Is(err, ethereum.NotFound) {
        continue // Transaction is still pending
    }
    log.Errorf("failed to get transaction receipt for hash %s: %+v", txHash.Hex(), err)
    return // Exits entire update loop
}
```

### Problem
The `return` statement on line 126 causes the entire update cycle to abort when any non-NotFound error occurs, leaving all subsequent transactions unprocessed.

## Resolution
Change the error handling to continue processing remaining transactions:

```go
receipt, err := mw.api.TransactionReceipt(ctx, txHash)
if err != nil {
    if errors.Is(err, ethereum.NotFound) {
[...]

PR #127, opened 2025-06-25
fix: handle API timeouts in MessageWatcherEth task

frrist:
- closes https://github.com/storacha/piri/issues/126


pkg/pdp/tasks/watcher_eth.go
```diff
@@ -112,70 +181,228 @@ func (mw *MessageWatcherEth) update() {
...
+				}
+			}
+
+			return nil
+		})
+	}
+
+	// Process results sequentially to avoid database conflicts
+	var updateCount int
+	wg := sync.WaitGroup{}
```
hannahhoward, line 224: nit: arguably a wait group for just one go routine is overkill (you can just make a done channel for example) 

non blocking

hannahhoward (approved):
Just to confirm, the switch to concurrent processing becomes generally neccesary as now you're doing retries and waiting longer?

exponential backoff has enough corner cases you could just use something like https://github.com/cenkalti/backoff

But also... nothing blocking here. LGTM

frrist:
> Just to confirm, the switch to concurrent processing becomes generally necessary as now you're doing retries and waiting longer

Yeah, basically. I want to ensure the db state is updated when waiting transactions complete as that unblocks dependent tasks, while also ensuring transactions that take a bit longer to fetch have the time to do so, w/o blocking other transactions.

PR #132, opened 2025-06-30
fix: don't require proofset and pdp url together

frrist:
- don't require flags together since values may come from the environment or a config file, in which case flags will be empty

### DIFF
--- a/cmd/cli/serve/ucan.go
+++ b/cmd/cli/serve/ucan.go
@@ -132,7 +132,7 @@ func startServer(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("loading config: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/cli/serve/ucan.go
```diff
@@ -133,6 +132,13 @@ func startServer(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("loading config: %w", err)
 	}
 
+	if len(cfg.PDPServerURL) > 0 && cfg.ProofSet == 0 {
```
volmedo, line 135: what about this small change for consistency in the checks?
```suggestion
	if cfg.PDPServerURL != "" && cfg.ProofSet == 0 {
```

volmedo (approved):

PR #135, opened 2025-07-01
fix: configurable presets from CI deployment

alanshaw:
Presets were recently changed and always ended up being for the warm staging network.

### DIFF
--- a/pkg/presets/presets.go
+++ b/pkg/presets/presets.go
@@ -17,32 +17,41 @@ var (
 	PrincipalMapping   map[string]string
 )
 
-// Setting this value to anything will cause production presets to be used, else use warm staging.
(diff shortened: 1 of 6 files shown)


pkg/presets/presets.go
```diff
@@ -17,32 +17,41 @@ var (
 	PrincipalMapping   map[string]string
 )
 
-// Setting this value to anything will cause production presets to be used, else use warm staging.
-var ProductionPresetsEnvVar = "PIRI_PRODUCTION_PRESETS"
```
alanshaw, line 21: Changed this to `PIRI_PRESETS` so that we don't have to have a different env var for each deployment. Technically a breaking change but actually it didn't work if you set it as there was no `else` or `return` on the conditional so the prod values got overwrittemn by warm staging values.

PR #138, opened 2025-07-01
Update docs and fix issues with CLI flag validation 

frrist:
## What
- Modifies the docs to provide a list of all required steps to run a Piri PDP and UCAN server
- Modifies CLI flags reducing validation for those with defaults.
- Guides readers to official Curio docs for PDP needs


cmd/cli/client/root.go
```diff
@@ -19,7 +19,6 @@ var (
 func init() {
 	Cmd.PersistentFlags().String("node-url", config.DefaultPDPClient.NodeURL, "URL of a Piri node")
 	cobra.CheckErr(viper.BindPFlag("node_url", Cmd.PersistentFlags().Lookup("node-url")))
-	cobra.CheckErr(Cmd.MarkPersistentFlagRequired("node-url"))
```
frrist, line 22: No longer require the flag as a working default is already present in the config

pkg/config/config.go
```diff
@@ -45,7 +45,7 @@ type UCANServer struct {
 
 	Port                    uint              `mapstructure:"port" validate:"required,min=1,max=65535" flag:"port"`
 	Host                    string            `mapstructure:"host" validate:"required" flag:"host"`
-	PDPServerURL            string            `mapstructure:"pdp_server_url" validate:"required,url" flag:"pdp-server-url"`
+	PDPServerURL            string            `mapstructure:"pdp_server_url" validate:"omitempty,url" flag:"pdp-server-url"`
```
frrist, line 48: This is an optional value, so remove the requirement it's present in the config and drop the default value as Piri can run a Ucan server w/o a PDP server.

PR #143, opened 2025-07-10
feat: resolve did:key from did:web

frrist:


pkg/principalresolver/httpresolver.go
```diff
@@ -0,0 +1,147 @@
...
+	for _, opt := range opts {
+		if err := opt(resolver); err != nil {
+			return nil, err
+		}
+	}
+	return resolver, nil
+}
+
+// TODO(forrest): the interface the implements in go-ucanto should probably accept a context
+// since means of resolution here are open ended, and may go to network or disk.
```
alanshaw, line 83: It does in new ucanto

cmd/cli/serve/ucan.go
```diff
@@ -313,9 +313,16 @@ func startServer(cmd *cobra.Command, _ []string) error {
 
 	defer svc.Close(ctx)
 
-	presolv, err := principalresolver.New(cfg.ServicePrincipalMapping)
+	presolv, err := principalresolver.NewHTTPResolver(map[string]string{
+		indexingServiceDID.String(): indexingServiceDID.String(),
+		uploadServiceDID.String():   uploadServiceDID.String(),
```
alanshaw, line 318: Not sure I understand why this is passed in here? Is this an allowlist?

I think the values are meant to be URL strings?

In theory the DID document you're meant to lookup is derived from the DID i.e. if you're asked to resolve `did:web:example.com`, you make a HTTP request to `https://example.com/.well-known/did.json`

pkg/presets/presets.go
```diff
@@ -61,7 +61,8 @@ var (
 	warmStageUploadServiceDID = lo.Must(did.Parse("did:web:staging.up.warm.storacha.network"))
 
 	warmStagePrincipalMapping = map[string]string{
-		warmStageUploadServiceDID.String(): "did:key:z6MkpR58oZpK7L3cdZZciKT25ynGro7RZm6boFouWQ7AzF7v",
+		warmStageUploadServiceDID.String():   "did:key:z6MkpR58oZpK7L3cdZZciKT25ynGro7RZm6boFouWQ7AzF7v",
+		warmStageIndexingServiceDID.String(): "did:key:z6Mkr4QkdinnXQmJ9JdnzwhcEjR8nMnuVPEwREyh9jp2Pb7k",
```
alanshaw, line 65: You don't need these anymore right?

alanshaw (changes requested):

pkg/presets/presets.go
frrist: Technically no, I suppose I could just delete the principal mappings completely.

cmd/cli/serve/ucan.go
frrist: I've made this change in 12f6449fac04e674b8ce5816c325ccb01e55f57c

alanshaw (approved):

PR #144, opened 2025-07-11
Implement (gradual) dependency injection via uber.fx

frrist:


frrist:
@hannahhoward I have address the duplicate code feedback in the following commits:
1. 5442ff5196a9b19d8d3ac0e2bb7ac235045880af removes the duplicate http handlers from the `fx` package, and instead wraps the existing handlers with echo.
2. 257b773393415a02348d0f0bcdc38231b4f3e2e4 remove the duplicate ucan handlers (service methods) from the `fx` package, now each service method is it's own function, provided as an option to the UCANServer constructor. Fairly mechanical change.

Note that I've added a new constructor for the claims service ([NewV2](https://github.com/storacha/piri/pull/144/files?diff=unified&w=0#diff-47db02effe15b1090d160201452704a19dbacbc7c38b107b16fc9c303551f197R52-R58)) that accepts a pre-configured publisher instead of creating one internally. This change separates publisher creation from claims service initialization, making the code easier to test and more flexible for different deployment scenarios.


If we prefer, I am happy to land 5442ff5196a9b19d8d3ac0e2bb7ac235045880af and 257b773393415a02348d0f0bcdc38231b4f3e2e4 as separate PRs, re-basing this one onto those.

pkg/service/claims/service.go
```diff
@@ -48,3 +49,10 @@ func New(id principal.Signer, claimStore claimstore.ClaimStore, publisherStore s
 
 	return &ClaimService{claimStore, publisher}, nil
 }
+
+func NewV2(
```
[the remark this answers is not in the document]
frrist: Yeah... I suppose I could have just renamed the original, was aiming to keep the diff small here. Eventually I suspect (hope) that the `New`/`Setup` method will be replaced in-favor of this new one in subsequent changes.

alanshaw (approved):

pkg/fx/config/provider.go
```diff
@@ -0,0 +1,291 @@
...
+		// Provide named values for services that still need them
+		fx.Annotate(
+			ProvideUploadServiceConnection,
+			fx.ResultTags(`name:"upload_service_connection"`),
+		),
+	),
+)
+
+// ProvideAppConfig provides the new organized application configuration
+func ProvideAppConfig(cfg config.UCANServer) (app.AppConfig, error) {
```
volmedo, line 39: I'd maybe call this `NewAppConfig` for consistency with other `New...` functions.

volmedo, line 39: also, this method is huge. I'd create functions to create each config and assemble the AppConfig from the different bits here.

pkg/fx/config/provider.go
```diff
@@ -0,0 +1,291 @@
...
+		Services: servicesConfig,
+	}, nil
+}
+
+// ProvideUploadServiceConnection provides the upload service connection for backward compatibility
+func ProvideUploadServiceConnection(cfg app.AppConfig) client.Connection {
+	return cfg.External.UploadService.Connection
+}
+
+func ReadPrivateKeyFromPEM(path string) (principal.Signer, error) {
```
volmedo, line 247: there is a duplicate of this function in the `cliutil` package that other packages use. Is that because these changes are an intermediate step and one of them will be removed once everything is consolidated?

Please remove it if that is not the case. If it needs to be kept, make it private, as it is only used within the package.

pkg/fx/storage/provider.go
```diff
@@ -0,0 +1,111 @@
...
+			fx.As(new(ucan.BlobAllocateService)),
+			fx.As(new(ucan.BlobAcceptService)),
+			fx.As(new(ucan.PDPInfoService)),
+			fx.As(new(ucan.ReplicaAllocateService)),
+		),
+	),
+)
+
+// StorageServiceParams contains all dependencies for the storage service
+type StorageServiceParams struct {
```
volmedo, line 34: since this type lives in the `storage` package, it could be called just `ServiceParams` or `Params`

pkg/fx/storage/provider.go
```diff
@@ -0,0 +1,111 @@
...
+	ID           principal.Signer
+	Blobs        *blobs.BlobService
+	Claims       *claims.ClaimService
+	PDP          pdp.PDP `optional:"true"`
+	ReceiptStore receiptstore.ReceiptStore
+	Replicator   *replicator.Service
+}
+
+// storageServiceWrapper wraps the storage service to implement the storage.Service interface
+type storageServiceWrapper struct {
```
volmedo, line 47: an alternative name would be `storageService` or just `service`

pkg/fx/storage/provider.go
```diff
@@ -0,0 +1,111 @@
...
+	blobs        *blobs.BlobService
+	claims       *claims.ClaimService
+	pdp          pdp.PDP
+	receiptStore receiptstore.ReceiptStore
+	replicator   *replicator.Service
+	uploadConn   client.Connection
+}
+
+// NewStorageService creates a new storage service
+func NewStorageService(params StorageServiceParams, lc fx.Lifecycle) (storage.Service, error) {
```
volmedo, line 58: alternatively, `NewService` would be consistent with the naming in other packages

pkg/fx/store/memory/provider.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewPublisherStore() store.FullStore {
+	ds := sync.MutexWrap(datastore.NewMapDatastore())
+	return store.FromDatastore(ds, store.WithMetadataContext(metadata.MetadataContext))
+}
+
+func NewReceiptStore() (receiptstore.ReceiptStore, error) {
+	ds := sync.MutexWrap(datastore.NewMapDatastore())
+	return receiptstore.NewDsReceiptStore(ds)
+}
```
volmedo, line 61: do all these `NewWhateverStore` need to be public?

The comment also applies to the definitions in the `fx/store/filesystem` package.

volmedo (approved):
LGTM.

I left some comments, mostly related to naming and visibility concerns. I'm missing some consistency when naming some types and constructor/provider functions.

PR #157, opened 2025-07-17
feat: implement telemetry package for metrics

frrist:
- timers
- guages
- counters
- info

### DIFF
--- a/pkg/telemetry/provider.go
+++ b/pkg/telemetry/provider.go
@@ -19,15 +19,16 @@ type Provider struct {
 }
 
 type Config struct {
-	ServiceName    string
-	ServiceVersion string
-	environment    string
-	endpoint       string
-	insecure       bool
-	headers        map[string]string
+	ServiceName     string
+	ServiceVersion  string
+	PublishInterval time.Duration
+	environment     string
+	endpoint        string
+	insecure        bool
+	headers         map[string]string
 }
 
-func NewProvider(ctx context.Context, cfg Config) (*Provider, error) {
+func newProvider(ctx context.Context, cfg Config) (*Provider, error) {
 	res, err := resource.New(ctx,
 		resource.WithAttributes(
 			attribute.String("service.name", cfg.ServiceName),
@@ -59,7 +60,7 @@ func NewProvider(ctx context.Context, cfg Config) (*Provider, error) {
 	provider := sdkmetric.NewMeterProvider(
 		sdkmetric.WithReader(
 			sdkmetric.NewPeriodicReader(exporter,
-				sdkmetric.WithInterval(30*time.Second),
+				sdkmetric.WithInterval(cfg.PublishInterval),
 			),
 		),
 		sdkmetric.WithResource(res),
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -3,6 +3,7 @@ package telemetry
 import (
 	"context"
 	"fmt"
+	"time"
 
 	logging "github.com/ipfs/go-log/v2"
 	"go.opentelemetry.io/otel/attribute"
@@ -12,8 +13,9 @@ import (
 var log = logging.Logger("telemetry")
 
 const (
-	defaultEndpoint    = "telemetry.storacha.network:443"
-	defaultEnvironment = "warm-staging"
+	defaultEndpoint        = "telemetry.storacha.network:443"
+	defaultEnvironment     = "warm-staging"
+	defaultPublishInterval = 30 * time.Second
 )
 
 type Telemetry struct {
@@ -25,8 +27,11 @@ func New(ctx context.Context, cfg Config) (*Telemetry, error) {
 	// collector endpoint and environment will be hard-coded for now
 	cfg.endpoint = defaultEndpoint
 	cfg.environment = defaultEnvironment
+	if cfg.PublishInterval == 0 {
+		cfg.PublishInterval = defaultPublishInterval
+	}
 
-	provider, err := NewProvider(ctx, cfg)
+	provider, err := newProvider(ctx, cfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create telemetry provider: %w", err)
 	}


pkg/telemetry/README.md
```diff
@@ -0,0 +1,313 @@
+# Telemetry Package
```
volmedo, line 1: I'd move all this content to a block comment in `telemetry.go` as this is the convention in Go.

pkg/telemetry/global.go
```diff
@@ -0,0 +1,124 @@
+package telemetry
```
volmedo, line 1: I think I'd simplify this changeset by removing this until there is a need for it

pkg/telemetry/info.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// Info represents an info metric - a gauge that always has value 1.0
+// Info metrics are used to expose textual information as labels.
+type Info struct {
+	gauge *FloatGauge
```
volmedo, line 13: why not a regular `Gauge` given that the value will always be `1`?

pkg/telemetry/info.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Update updates the info metric with new label values
+func (i *Info) Update(ctx context.Context, labels map[string]string) {
+	attrs := make([]attribute.KeyValue, 0, len(labels))
+	for k, v := range labels {
+		attrs = append(attrs, attribute.String(k, v))
+	}
+	i.attrs = attrs
+	i.Record(ctx)
+}
```
volmedo, line 60: since this is an `Info` metric, I think the most common way of using it will be updating the value of a single label. As it is, `Record` is kinda useless and `Update` requires passing all the labels. I think a better API would be having a single `Record` method that accepts a map of labels and then merges it with the one already present in the `Info`.

pkg/telemetry/info.go
```diff
@@ -0,0 +1,135 @@
...
+
+// ConstantGauge represents a gauge metric with a constant value
+// This is useful for configuration values that don't change often
+type ConstantGauge struct {
+	gauge *FloatGauge
+	value float64
+	attrs []attribute.KeyValue
+}
```
volmedo, line 68: I don't think this type is very useful and I'd remove it

pkg/telemetry/info.go
```diff
@@ -0,0 +1,135 @@
...
+// UpdateValue updates the constant value and records it
+func (c *ConstantGauge) UpdateValue(ctx context.Context, value float64) {
+	c.value = value
+	c.Record(ctx)
+}
+
+// InfoRegistry helps manage multiple info metrics
+type InfoRegistry struct {
+	infos map[string]*Info
+}
```
volmedo, line 116: same here, I'd add this once there is a need for it

volmedo (commented):
@frrist some comments from a quick review. Let me know what you think. I'll implement them myself if you don't get to this before.

pkg/telemetry/README.md
frrist: Love it

pkg/telemetry/global.go
frrist: My gut check says using the global will be easier from an implementation perspective, else I _think_ we'll need to plumb an instance of telemetry into each service that records telemetry. Though I ack we aren't at that point yet, as we are mainly looking to record just the node version and DID. Will leave it up to you: happy to remove this, happy to leave it around.

pkg/telemetry/info.go
frrist: Yup, on second look this adds zero value - let's nix it.

pkg/telemetry/info.go
frrist: Agreed, let's nix it too.

pkg/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,724 @@
+package telemetry_test
```
frrist, line 1: I'm not sure these tests add a lot of value at the moment, happy to remove them too if it makes implementation faster.

pkg/telemetry/info.go
volmedo: I'll leave it open for now, but that's a good point to consider if we keep using it with the same labels again and again.

pkg/telemetry/telemetry.go
```diff
@@ -0,0 +1,272 @@
...
+type Telemetry struct {
+	provider *Provider
+	meter    metric.Meter
+}
+
+func New(ctx context.Context, cfg Config) (*Telemetry, error) {
+	// collector endpoint and environment will be hard-coded for now
+	cfg.endpoint = defaultEndpoint
+	cfg.environment = defaultEnvironment
+	cfg.insecure = true
```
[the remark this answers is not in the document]
frrist: also, ack-ing these are my changes I am reviewing, so comments are all critiques of my mistakes.

frrist (commented):
can't approve my own PR, but looks good once comments are addressed.

pkg/telemetry/provider.go
```diff
@@ -0,0 +1,82 @@
...
+
+	exporter, err := otlpmetricgrpc.New(ctx, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create metric exporter: %w", err)
+	}
+
+	provider := sdkmetric.NewMeterProvider(
+		sdkmetric.WithReader(
+			sdkmetric.NewPeriodicReader(exporter,
+				sdkmetric.WithInterval(30*time.Second),
```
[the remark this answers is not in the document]
frrist: address in latest commit

cmd/cli/root.go
```diff
@@ -80,3 +86,19 @@ func initConfig() {
 		cobra.CheckErr(viper.ReadInConfig())
 	}
 }
+
+func initTelemetry() {
+	// bail if this has been disabled.
+	if os.Getenv("PIRI_DISABLE_ANALYTICS") != "" {
+		return
+	}
```
frrist, line 94: Will need to document this opt-out procedure, and also include a list of the data we collect when it's enabled.

pkg/telemetry/counter.go
```diff
@@ -0,0 +1,111 @@
...
+	allAttrs := append(c.attrs, attrs...)
+	c.counter.Add(ctx, value, metric.WithAttributes(allAttrs...))
+}
+
+func (c *FloatCounter) WithAttributes(attrs ...attribute.KeyValue) *FloatCounter {
+	return &FloatCounter{
+		counter: c.counter,
+		attrs:   append(c.attrs, attrs...),
+	}
+}
```
alanshaw, line 111: Seems wild to have to implement all this? Is there not some library code we can use?

alanshaw (approved):
LGTM

pkg/telemetry/telemetry.go
```diff
@@ -1,172 +1,18 @@
...
-//	telemetry.IntAttr("status_code", 200)
-//	telemetry.BoolAttr("success", true)
-//
-// # Predefined Boundaries
-//
-// The package provides predefined boundary values for common use cases:
-//
-//	telemetry.DefaultBoundaries  // General purpose boundaries
-//	telemetry.LatencyBoundaries  // For request latencies (seconds)
-//	telemetry.SizeBoundaries     // For response/request sizes (bytes)
```
volmedo: I agree some of these comments were a bit generic and described counters, gauges and such, but some others were specific to how this package is supposed to be used. I would've kept at least those, such as the options to use the package (component-specific telemetry instances vs. global instance) as well as the examples.

pkg/telemetry/provider.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp"
```
volmedo: why are we switching to HTTP? gRPC is supposed to be more efficient and it's the native alternative in Go.

pkg/telemetry/telemetry.go
```diff
@@ -179,7 +25,6 @@ func New(ctx context.Context, cfg Config) (*Telemetry, error) {
 	// collector endpoint and environment will be hard-coded for now
 	cfg.endpoint = defaultEndpoint
 	cfg.environment = defaultEnvironment
-	cfg.insecure = true
```
volmedo: ❤️

pkg/telemetry/counter.go
frrist: They are for convenience, we don't _have_ to implement all of these methods and _could_ just call the otel package methods directly.

pkg/telemetry/telemetry.go
frrist: My preference would be for dedicated comments on each telemetry method, I'd like to avoid blocking this merge on that, okay if we land in a follow on?

pkg/telemetry/provider.go
volmedo: I agree on using a port different from 443 😄. Why not use the default 4317? I'd argue that is even less surprising, since it's makes clear what is being sent is OTLP telemetry data.

I'm not sure about the simplicity argument. It will be piri instances talking to the server, so I don't think there is a need to optimize for compatibility (at least for now). Also, given the nature of the data I'm not sure we will be using curl to troubleshoot either.

Of course we can add it later, but it looks like it is actually less (or same) effort to use gRPC from the get go. Also, deprecating something (OTLP over HTTP in this case) later when a lot of servers are using it will be difficult.

PR #159, opened 2025-07-17
feat: server info as JSON

alanshaw:
e.g.

```sh
curl -H "Accept: application/json" https://piri.example.com/
{"id":"did:key:z6Mkmcc4QaxDQJaGMD63H3sg5VjyMY6HyBd4H1HEJ5yYMcii","build":{"version":"v0.0.10-94fee20-dirty","repo":"https://github.com/storacha/piri"}}
```


pkg/server/server.go
```diff
@@ -76,11 +78,38 @@ func NewServer(service storage.Service, options ...server.Option) (*http.ServeMu
...
+	}
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Write([]byte(fmt.Sprintf("🔥 piri %s\n", build.Version)))
-		w.Write([]byte("- https://github.com/storacha/piri\n"))
-		w.Write([]byte(fmt.Sprintf("- %s", id.DID())))
+		if strings.Contains(r.Header.Get("Accept"), "application/json") {
+			json, err := json.Marshal(&info)
+			if err != nil {
+				log.Errorf("failed JSON marshal server info: %w", err)
+				http.Error(w, "failed JSON marshal server info", 500)
```
frrist, line 105: nit: would prefer to use the dedicated http constant for internal server error instead of a "magic" number:
```suggestion
                                http.Error(w, "failed JSON marshal server info", http.StatusInternalServerError)
```

pkg/server/server.go
```diff
@@ -76,11 +78,38 @@ func NewServer(service storage.Service, options ...server.Option) (*http.ServeMu
...
+				return
+			}
+			w.Write(json)
+		} else {
+			w.Write([]byte(fmt.Sprintf("🔥 piri %s\n", info.Build.Version)))
+			w.Write([]byte("- https://github.com/storacha/piri\n"))
+			w.Write([]byte(fmt.Sprintf("- %s", info.ID)))
```
frrist, line 112: nit: may also be worth setting the content type here when not returning json, though probably not all that important.
```suggestion
   		        w.Header().Set("Content-Type", "text/plain; charset=utf-8")
			w.Write([]byte(fmt.Sprintf("🔥 piri %s\n", info.Build.Version)))
			w.Write([]byte("- https://github.com/storacha/piri\n"))
			w.Write([]byte(fmt.Sprintf("- %s", info.ID))
```

pkg/server/server.go
```diff
@@ -76,11 +78,38 @@ func NewServer(service storage.Service, options ...server.Option) (*http.ServeMu
 	return mux, nil
 }
 
+type ServerInfo struct {
+	ID    string    `json:"id"`
```
frrist, line 82: nit: worth calling this `DID` instead?

frrist (approved):
Good stuff! Bunch of nits, address them if you think it's worth it, or merge as is. Approval is sticky.

pkg/server/server.go
alanshaw: I would prefer ID - this is the server identity (it's "id"), the type of the ID happens to be DID...

PR #160, opened 2025-07-17
refactor: move ucan service methods to own package

frrist:
mechanical change, extracted from https://github.com/storacha/piri/pull/144


pkg/service/storage/ucan/blob_accept.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+type BlobAcceptService interface {
+	ID() principal.Signer
+	PDP() pdp.PDP
+	Blobs() blobs.Blobs
+	Claims() claims.Claims
+}
+
+func BlobAccept(storageService BlobAcceptService) server.Option {
```
volmedo, line 27: very minor preference nitpick, feel free to disregard: add a `With` prefix to the method name to make it a bit clearer it returns an option, a `Method` suffix (or even `ServiceMethod`, but that might be too verbose) because it returns, more specifically, a `ServiceMethod` option, and use a more generic name for the argument.

E.g.
```suggestion
func WithBlobAcceptMethod(service BlobAcceptService) server.Option {
```

and equivalent changes in the other functions.

volmedo (approved):
LGTM, left a non-blocking comment.

PR #163, opened 2025-07-30
fix: chain scheduler robustness

frrist:
ensures the chain notify channel continues to send notification, if one is not sent with 60s we restart the chain notify channel from lotus

- ported from https://github.com/filecoin-project/curio/pull/552


frrist:
Yes - these changes are copy pasted directly from Curio. Would have imported it directly if it didn't depend on curio specific build flags :disappointed:

PR #164, opened 2025-07-30
fix: add timeout to eth api requests to prevent infinite hanging

frrist:
- similiar fix to change in https://github.com/filecoin-project/curio/pull/553

### DIFF
--- a/pkg/pdp/tasks/watcher_eth.go
+++ b/pkg/pdp/tasks/watcher_eth.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/cenkalti/backoff/v5"
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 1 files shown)


pkg/pdp/tasks/watcher_eth.go
```diff
@@ -275,25 +282,31 @@ func (mw *MessageWatcherEth) checkTransaction(ctx context.Context, txHash common
...
+		defer cancel()
+
+		r, err := mw.api.TransactionReceipt(txCtx, txHash)
 		// Don't retry on NotFound errors
 		if errors.Is(err, ethereum.NotFound) {
 			return nil, backoff.Permanent(err)
 		}
 		return r, err
-	}, backoff.WithMaxTries(mw.maxEthAPIRetries), backoff.WithBackOff(backoff.NewExponentialBackOff()))
+	}, backoff.WithMaxTries(mw.maxEthAPIRetries), backoff.WithBackOff(backoff.NewConstantBackOff(time.Second)))
```
hannahhoward, line 294: nb: are we sure we want constant backoff here?

hannahhoward (approved):

pkg/pdp/tasks/watcher_eth.go
frrist: Fairly cretin, yes. Moved to a constant backoff to make it easier to reason about: The method will retry a maximum of 10 times, waiting one second between each failed attempt. The Lotus API will have three seconds to reply before we timeout the request due to cancellation.

PR #170, opened 2025-08-02
refactor(testutil): use external testutil library

hannahhoward:
# Goals

refactor based on https://github.com/storacha/go-libstoracha/pull/40 -- should not be merged until that PR is merged and tagged

# Implementation

- use go-libstoracha's testutil library, instead of having a local internal one
- there were some extra functions in this testutil, but they weren't used at all
- the libstoracha piece func was called RandomPiece, so changed references to CreatePiece


alanshaw (approved):
Feels good 👏

Issue #97, opened 2025-06-10
Implement Dynamic Logging Control Command


frrist:
## Description

The PDP server needs a logging command similar to Lotus's implementation that allows operators to list logging subsystems and dynamically change their log levels at runtime. This will help with debugging and monitoring without requiring server restarts.

## Current Behavior

- The codebase uses `github.com/ipfs/go-log/v2` for logging, which makes this an easy change.
- Log levels are set at compile time or through environment variables
- No runtime control over logging subsystems
- 29 files use the logging system with various subsystem names like "pdp-auth", "pdp/aggregator", "piri", etc.
- No way to discover available logging subsystems at runtime

## Expected Behavior

Implement a `piri log` command that provides:
1. List all available logging subsystems
2. Get current log level for a subsystem
3. Set log level for specific subsystems at runtime
4. Set log level for all subsystems
5. Changes persist until the process restarts
  - Future enhancement, persist changes after restart via dynamic config: https://github.com/storacha/piri/issues/93

### Reference Implementation

Lotus's logging command (https://github.com/filecoin-project/lotus/blob/master/cli/log.go) provides a good model:
```bash
# List all subsystems and their levels
lotus log list

# Set log level for a subsystem
lotus log set-level <subsystem> <level>

# Set log level for all subsystems
lotus log set-level --all <level>
```

### Implementation Requirements

1. **Create Logging Command** (`cmd/log.go`):
```go
package cmd

import (
    "fmt"
    logging "github.com/ipfs/go-log/v2"
    "github.com/urfave/cli/v2"
)

var LogCmd = &cli.Command{
    Name:  "log",
    Usage: "Manage logging subsystems and levels",
    Subcommands: []*cli.Command{
        logListCmd,
        logSetLevelCmd,
    },
}

var logListCmd = &cli.Command{
    Name:  "list",
    Usage: "List all logging subsystems and their levels",
    Action: func(c *cli.Context) error {
        // List all registered subsystems
        subsystems := logging.GetSubsystems()
        for _, subsystem := range subsystems {
            level := logging.GetLevel(subsystem)
            fmt.Printf("%-30s %s\n", subsystem, level)
        }
        return nil
    },
}

var logSetLevelCmd = &cli.Command{
    Name:      "set-level",
    Usage:     "Set log level for a subsystem",
    ArgsUsage: "<subsystem> <level>",
    Flags: []cli.Flag{
        &cli.BoolFlag{
            Name:  "all",
[...]

PR #173, opened 2025-08-06
feat(logging): add dynamic logging CLI and management API

bomanaps:
This Pr closes this https://github.com/storacha/piri/issues/97
I have:

   * Created a new log command with list and set-level subcommands.
   * Created a management API server that exposes endpoints for listing and setting
      log levels.
   * Integrated the management server into the main application.
   * Used the cobra library to create the CLI commands.
   * Used the go-log/v2 library to manage logging.

  The log command can be used like this:
   * piri log list: Lists all logging subsystems and their current levels.
   * piri log set-level <subsystem> <level>: Sets the log level for a specific
     subsystem.
   * piri log set-level --all <level>: Sets the log level for all subsystems.
Please Note this is still a WIP and I still have some missing uint test implementation

### DIFF
--- a/cmd/cli/log.go
+++ b/cmd/cli/log.go
@@ -1,0 +1,69 @@
+package cli
+
+import (
+	"fmt"
+
+	logging "github.com/ipfs/go-log/v2"
+	"github.com/spf13/cobra"
+	"github.com/storacha/piri/cmd/cli/client"
+)
+
+var logListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List all logging subsystems and their levels",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// List all registered subsystems
+		subsystems := logging.GetSubsystems()
+		for _, subsystem := range subsystems {
+			level := logging.Logger(subsystem).Level().String()
+			fmt.Printf("%-30s %s\n", subsystem, level)
+		}
+		return nil
+	},
+}
+
+var logSetLevelCmd = &cobra.Command{
+	Use:   "set-level <subsystem> <level>",
+	Short: "Set log level for a subsystem",
+	Args: func(cmd *cobra.Command, args []string) error {
+		all, err := cmd.Flags().GetBool("all")
+		if err != nil {
+			return err
+		}
+		if all {
+			return cobra.ExactArgs(1)(cmd, args)
+		}
+		return cobra.ExactArgs(2)(cmd, args)
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		all, err := cmd.Flags().GetBool("all")
+		if err != nil {
+			return err
+		}
+		if all {
+			level := args[0]
+			// get all subsystems
+			subsystems := logging.GetSubsystems()
+			for _, subsystem := range subsystems {
+				if err := client.SetLogLevel(cmd.Context(), subsystem, level); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+		subsystem := args[0]
+		level := args[1]
+		return client.SetLogLevel(cmd.Context(), subsystem, level)
+	},
+}
+
+func init() {
+	logSetLevelCmd.Flags().Bool("all", false, "Set level for all subsystems")
+	LogCmd.AddCommand(logListCmd)
+	LogCmd.AddCommand(logSetLevelCmd)
+}
+
+var LogCmd = &cobra.Command{
+	Use:   "log",
+	Short: "Manage logging subsystems and levels",
+}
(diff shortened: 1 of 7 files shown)


cmd/cli/log.go
```diff
@@ -0,0 +1,69 @@
...
+		subsystems := logging.GetSubsystems()
+		for _, subsystem := range subsystems {
+			level := logging.Logger(subsystem).Level().String()
+			fmt.Printf("%-30s %s\n", subsystem, level)
+		}
+		return nil
+	},
+}
+
+var logSetLevelCmd = &cobra.Command{
```
frrist, line 25: Actually, I've been thinking about this more and want to suggest a different approach:

`piri log set-level <level>` would set the level across all logging systems, while `piri log set-level --system <system1> --system <system2> <level>` would target specific systems. You can pass multiple `--system` flags to configure several systems at once.
This feels cleaner and gives users the flexibility to be either broad or precise depending on what they need. For the implementation, `StringSlice` would work better than `StringArray` for the flag handling.
What do you think about this direction?

cmd/cli/log.go
```diff
@@ -0,0 +1,69 @@
...
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		all, err := cmd.Flags().GetBool("all")
+		if err != nil {
+			return err
+		}
+		if all {
+			level := args[0]
+			// get all subsystems
+			subsystems := logging.GetSubsystems()
```
frrist, line 46: similar comment here https://github.com/storacha/piri/pull/173/files#r2257590145

cmd/cli/log.go
```diff
@@ -0,0 +1,69 @@
...
+func init() {
+	logSetLevelCmd.Flags().Bool("all", false, "Set level for all subsystems")
+	LogCmd.AddCommand(logListCmd)
+	LogCmd.AddCommand(logSetLevelCmd)
+}
+
+var LogCmd = &cobra.Command{
+	Use:   "log",
+	Short: "Manage logging subsystems and levels",
+}
```
frrist, line 69: nit: style thing - please move these to top of file, i.e.:
```go
// define root level command
var LogCmd ...

// add sub commands and define flags
func init(){...}

// define sub commands
var logListCmd = ...
var logSetLeveCmd =  ...
```

pkg/management/server.go
```diff
@@ -0,0 +1,46 @@
...
+	if req.Level == "" {
+		return c.String(http.StatusBadRequest, "level is required")
+	}
+
+	if err := logging.SetLogLevel(req.Subsystem, req.Level); err != nil {
+		return c.String(http.StatusBadRequest, err.Error())
+	}
+
+	return c.NoContent(http.StatusOK)
+}
```
frrist, line 46: I'm actually having second thoughts about making this its own server. Instead, let's build it as an admin API that we can register on the existing PDP and UCAN servers.

We could create a `RegisterAdminRoutes` method (or something similar) that takes an echo server instance. Then we'd call this method in both:

https://github.com/storacha/piri/blob/main/pkg/pdp/api/server.go#L43
https://github.com/storacha/piri/blob/main/pkg/server/server.go#L44

While we're at it, let's move this to `pkg/admin/api.go` to better reflect what it is.

Also, we should define proper Request and Response types that both the CLI and API can use, rather than these inline struct definitions:
```go
var req struct {
    Subsystem string `json:"subsystem"`
    Level     string `json:"level"`
}
```

This approach keeps things cleaner and gives us better reusability across the codebase.

cmd/cli/root.go
```diff
@@ -62,12 +62,16 @@ func init() {
 	cobra.CheckErr(rootCmd.MarkPersistentFlagFilename("key-file", "pem"))
 	cobra.CheckErr(viper.BindPFlag("key_file", rootCmd.PersistentFlags().Lookup("key-file")))
 
+	rootCmd.PersistentFlags().String("manage-api", "127.0.0.1:8888", "Management API address")
+	cobra.CheckErr(viper.BindPFlag("manage_api", rootCmd.PersistentFlags().Lookup("manage-api")))
+
```
frrist, line 67: Per https://github.com/storacha/piri/pull/173/files#r2257733914 we can remove this.

frrist (commented):
Good stuff! Thank you @bomanaps <3

Sorry for the churn here, but I think it's best if we deviate from the initial issue - as will be evident from a few of the comments. Let me know if you have any questions.

cmd/cli/client/pdp/piece/info.go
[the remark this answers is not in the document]
bomanaps: sorry i had incorrect assumption, my understanding at that time was that the cmd/cli/client directory only contained the client-side implementation related to the logging feature, making it redundant after the refactoring I did.

frrist (commented):
looking good, few more comments

bomanaps:
Hey @frrist please can you trigger the ci again although am not sure how to go about the AWS fail

frrist (approved):

PR #177, opened 2025-08-07
fix: normalize public url to remove any trailing slashes

frrist:


alanshaw (approved):

frrist:
going with https://github.com/storacha/piri/pull/176 instead as this is overkill.

PR #180, opened 2025-08-08
fix: join HTTP path to multiaddr

alanshaw:
This joins a HTTP path to a multiaddr. The multiaddr is allowed to already have an existing HTTP path. If it has an existing path it is joined to the end of it, and if it does not have an existing HTTP path then the "http-path" protocol is added to the multiaddr with the specified value.


volmedo (approved):
💯 

I wonder if this should be done in `multiaddr.Multiaddr.Join`, or even just returning an error if a `http-path` component already exists. It's unexpected that `Join` produces an invalid multiaddress IMO.

alanshaw:
> I wonder if this should be done in `multiaddr.Multiaddr.Join`, or even just returning an error if a `http-path` component already exists. It's unexpected that `Join` produces an invalid multiaddress IMO.

I think it's a bit specific to the `http-path` protocol to be in the multiaddr library but I agree `Join` should return an error.

PR #181, opened 2025-08-11
fix: indexer UCAN endpoint

alanshaw:
The UCAN endpoint is at `/claims` not `/`. For some reason in a previous deployment of the indexer this worked anyways.


volmedo (approved):
💯

PR #182, opened 2025-08-11
fix: joining blob path to PDP public URL

alanshaw:
Note, targets: https://github.com/storacha/piri/pull/180

* Moves `joinHTTPPath` to a package that can be used everywhere.
* Uses it to join the PDP `piece/{blobCID}` to the public curio multiaddr.
* Removes the `TrimSuffix` from [#176](https://github.com/storacha/piri/pull/176) as it is not needed.

### DIFF
--- a/pkg/mahttp/joinpath.go
+++ b/pkg/mahttp/joinpath.go
@@ -1,0 +1,46 @@
+package mahttp
+
+import (
+	"net/url"
+
+	"github.com/multiformats/go-multiaddr"
+)
+
+// JoinPath joins a HTTP path onto an existing multiaddr. If the multiaddr
+// includes the "http-path" protocol already, it is joined to the end of the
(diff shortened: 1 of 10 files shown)


pkg/mahttp/joinpath.go
```diff
@@ -0,0 +1,46 @@
+package mahttp
```
frrist, line 1: How would we feel about putting this in `lib/` instead of it's own `mahttp` package?

frrist (approved):

pkg/mahttp/joinpath.go
frrist: Whoops, you're right one doesn't exist yet. Would you mind creating one and moving this logic there? Over in https://github.com/storacha/piri/pull/169 I have done something similar - defined a `lib/` dir and started moving common logic over to it.

PR #186, opened 2025-08-11
feat: implement `proofset list` cli/client/api

frrist:
allows users to list the proofsets their node is currently configured with. This will come in handy w/ later PRs where we init a node and a proofset automatically if one doesn't exist when a "full" node starts. The intention here is to use this API on startup in a later PR.

### DIFF
--- a/cmd/cli/client/pdp/proofset/list.go
+++ b/cmd/cli/client/pdp/proofset/list.go
@@ -1,0 +1,46 @@
+package proofset
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/storacha/piri/pkg/config"
+	"github.com/storacha/piri/pkg/pdp/httpapi/client"
+)
+
+var (
+	ListCmd = &cobra.Command{
+		Use:   "list",
+		Short: "List all proof sets",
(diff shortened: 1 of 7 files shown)


cmd/cli/client/pdp/proofset/list.go
```diff
@@ -0,0 +1,46 @@
...
+	api, err := client.NewFromConfig(cfg)
+	if err != nil {
+		return fmt.Errorf("creating client: %w", err)
+	}
+
+	proofSet, err := api.ListProofSet(ctx)
+	if err != nil {
+		return fmt.Errorf("getting proof set status: %w", err)
+	}
+	jsonProofSet, err := json.MarshalIndent(proofSet, "", "  ")
```
alanshaw, line 39: Is there a more friendly format that can be output by default? I feel like you should request JSON output if that's what you want.

alanshaw (approved):
LGTM but left a suggestion about output format...

cmd/cli/client/pdp/proofset/list.go
frrist: Yeah this can be improved to format the output as a table, for example. Treating that as out of scope for now.

PR #187, opened 2025-08-11
Frrist/refactor/task engine start

frrist:
Adds dedicated start and stop methods (lifecycle) to task engine and its respective tasks.


volmedo (approved):
LGTM as a refactor.

I got a couple questions about the original implementation though:
- do we limit the concurrency in any way?
- it doesn't look like the engine waits for ongoing tasks to finish before returning from `Stop`. Should it block until in-flight work is finished?

PR #192, opened 2025-08-13
Single Process Piri Node

frrist:
# Dependencies & Review
This PR builds on https://github.com/storacha/piri/pull/190
I would recommend reviewing each commit of this PR on it's own - I have done my best to keep them well scoped.

# What
This PR makes life easier by combining two servers into one. Instead of juggling `piri serve ucan` and `piri server pdp` in separate processes, you can now just run `piri serve full` and call it a day.

The new command runs both UCAN and PDP servers in a single process. Don't worry if you're using an external PDP implementation (like Curio) - `piri serve ucan` still works standalone.

Here's what the config looks like - notice how both UCAN and PDP settings live together:
```toml
[identity]
key_file = '/etc/piri/service.pem'

[pdp]
contract_address = '0x6170dE2b09b404776197485F3dc6c968Ef948505'
lotus_endpoint = 'wss://lotus.spicystorage.tech/rpc/v1'
owner_address = '0x7469B47e006D0660aB92AE560b27A1075EEcF97F'

[repo]
data_dir = '/data/piri'
temp_dir = '/tmp/storage'

[server]
host = 'localhost'
port = '3000'
public_url = 'https://piri.spicystorage.tech'

[ucan]
proof_set = '414'

[ucan.services]
[ucan.services.indexer]
did = 'did:web:staging.indexer.warm.storacha.network'
url = 'https://staging.indexer.warm.storacha.network/claims'
proof = '<proof>'

[ucan.services.principal_mapping]
'did:web:staging.indexer.warm.storacha.network' = 'did:key:z6Mkr4QkdinnXQmJ9JdnzwhcEjR8nMnuVPEwREyh9jp2Pb7k'
'did:web:staging.up.warm.storacha.network' = 'did:key:z6MkpR58oZpK7L3cdZZciKT25ynGro7RZm6boFouWQ7AzF7v'

[ucan.services.publisher]
ipni_announce_urls = ['https://cid.contact/announce', 'https://staging.ipni.warm.storacha.network']

[ucan.services.upload]
did = 'did:web:staging.up.warm.storacha.network'
url = 'https://staging.up.warm.storacha.network'
``` 

# How
This brings together work from several previous PRs (#187, #186, #169, #161, #144) to let both UCAN and PDP modules run in the same Piri process. Here's what changed under the hood:

**Architecture Changes:**
- Refactored the config system to be composable - configs are now split into logical sections (identity, server, repo, ucan, pdp) that can be mixed and matched
[...]

PR #196, opened 2025-08-21
fix: explicitly log handler execution errors

alanshaw:
Unexpected errors from handlers are sent to this function, [by default it prints to stderr](https://github.com/storacha/go-ucanto/blob/cd24d1a476ed38d5a70d443c60e0f11b361ff1f7/server/server.go#L106). This PR explicitly logs the error using our logger and prints the stack trace if the underlying cause had a stack trace.

### DIFF
--- a/cmd/cli/serve/ucan.go
+++ b/cmd/cli/serve/ucan.go
@@ -365,6 +365,13 @@ func startServer(cmd *cobra.Command, _ []string) error {
 		fmt.Sprintf("%s:%d", cfg.Server.Host, cfg.Server.Port),
 		svc,
 		ucanserver.WithPrincipalResolver(cachedpresolv.ResolveDIDKey),
+		ucanserver.WithErrorHandler(func(err ucanserver.HandlerExecutionError[any]) {
+			log.Error(err.Error())
+			stack := err.Stack()
+			if stack != "" {
+				log.Error(stack)
+			}
+		}),
 	)
 	return err
 
--- a/cmd/cli/serve/ucan.go
+++ b/cmd/cli/serve/ucan.go
@@ -366,11 +366,11 @@ func startServer(cmd *cobra.Command, _ []string) error {
 		svc,
 		ucanserver.WithPrincipalResolver(cachedpresolv.ResolveDIDKey),
 		ucanserver.WithErrorHandler(func(err ucanserver.HandlerExecutionError[any]) {
-			log.Error(err.Error())
-			stack := err.Stack()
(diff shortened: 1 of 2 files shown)


volmedo (approved):
❤️

cmd/cli/serve/ucan.go
```diff
@@ -365,6 +365,13 @@ func startServer(cmd *cobra.Command, _ []string) error {
 		fmt.Sprintf("%s:%d", cfg.Server.Host, cfg.Server.Port),
 		svc,
 		ucanserver.WithPrincipalResolver(cachedpresolv.ResolveDIDKey),
+		ucanserver.WithErrorHandler(func(err ucanserver.HandlerExecutionError[any]) {
+			log.Error(err.Error())
+			stack := err.Stack()
+			if stack != "" {
+				log.Error(stack)
+			}
+		}),
```
frrist, line 374: Can you also append this option here: https://github.com/storacha/piri/blob/main/pkg/fx/ucan/provider.go#L41 - this will ensure its enabled on the full server. Will merge https://github.com/storacha/piri/pull/192 after this lands and manage conflicts in #192.

Also, would prefer to log the stack and error as a single log line, including an error message like:
```go
l := log.With("error", err.Error()
if s := err.Stack(); s != nil {
  l.With("stack", s)
}
l.Error("ucan handler execution error")
```

Alternatively, and maybe for a future change, the type `ucanserver.HandlerExecutionError` could implement the [`MarshalLogObject`](https://pkg.go.dev/go.uber.org/zap#Objects) interface, and then be passed directly to a logging call, e.g.
```go
type handlerErrorLogMarshaller struct {
	err ucanserver.HandlerExecutionError[any]
}

func (h handlerErrorLogMarshaller) MarshalLogObject(e zapcore.ObjectEncoder) error {
	if h.err == nil {
		return nil
	}
	e.AddString("error", h.err.Error())
	if h.err.Name() != "" {
		e.AddString("name", h.err.Name())
	}
	if h.err.Stack() != "" {
		e.AddString("stack", h.err.Stack())
	}
	if h.err.Cause() != nil {
		e.AddString("cause", h.err.Cause().Error())
	}
	if caps := h.err.Capability(); caps != nil {
		capsb, err := caps.MarshalJSON()
		if err != nil {
			e.AddString("capability_unmarshal_error", err.Error())
		} else {
			e.AddString("capability", string(capsb))
		}
	}
	return nil
}
```

frrist (approved):

PR #203, opened 2025-08-27
Frrist/docs/single process

frrist:


hannahhoward (approved):
LGTM.

One comment: Am I correct that the init command generates a configuration, which must be copy/pasted to a file, and then passed to `piri serve full`?

Is there a reason not just write the config somewhere automatically in the init command, and then assume the configuration is there by default on `piri serve full`

(i.e. ~/.piri/config.toml or something)

non blocking but a suggestion for next version.

PR #216, opened 2025-09-03
fix: make replicate Transfer idempotent

### LINKED ISSUE #215: Replica Transfer Request Failure Cases
### Discussed in https://github.com/storacha/piri/discussions/214

<div type='discussions-op-text'>

<sup>Originally posted by **frrist** September  3, 2025</sup>
When Piri receives a UCAN invocation to allocate a replica, here is what happens:

1. A transfer invocation and the node creates an async task to allocate a replica:
https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/storage/ucan/replica_allocate.go#L128-L137

2. The transfer request is enqueued into a replication job queue:
https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/replicator/replicate.go#L86-L88
    2.a. Note, this queue is configured to re-try the request up to 10 times before giving up: https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/replicator/replicate.go#L63
    2.b. This queue will allow as many parallel transfer requests as the node has CPUs: https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/replicator/replicate.go#L64

3. When the the node has capacity to execute the transfer request the `Transfer` method is called:
https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/replicator/replicate.go#L69-L82

4. The transfer method pulls the data from the "Source" node it is replicating the data from:
https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/storage/handlers/replica/transfer.go#L135-L172
    4.a. If any part of this transfer fails fails due to network issues, `Transfer` returns an error, requeueing the transfer task. This is good, we want to make sure we pull the data we are supposed to replicate.

5. Once the transfer of data is complete, the node creates a receipt for the transfer invocation, and fires it off to the upload service: https://github.com/storacha/piri/blob/c50ea6b0b818f8a4d6c500b3cb59248ab136c618/pkg/service/storage/handlers/replica/transfer.go#L174-L235

Problems arise at step 5. 
Once the node successfully performs the transfer, and accepts (blob/accept) the piece is now stored in Piri. However, if a failure occurs while accepting the blob, or issuing a receipt to the upload service, and error is returned, and the entire transfer operation starts again. This can lead to multiple redundant transfers occurring when failures post-transfer occur. We actually saw this play out in our replication tests, some nodes repeatedly allocated the same blob multiple times when communicating with the upload service.
This is a classic idempotency problem. The transfer operation could instead be made idempotent. That way, if the data has already been transferred successfully, we shouldn't transfer it again just because the receipt delivery failed, or the blob accept operation failed. 

I think there are (at least) two issue here:
1. `blobhandler.Accept` is not idempotent, when called `N` times with the same blob it will create `N` allocations - I believe this is intended by design.
2. Transfers (moving the bytes between nodes) may occur multiple times during failure cases.

Together these issues compound into a larger problem, e.g.
1. First attempt: Transfer succeeds, Accept succeeds, but receipt sending fails
2. Retry:
    - Transfer happens AGAIN (wasteful but succeeds)
    - Accept is called AGAIN (might fail or create duplicates)
    - Receipt sending is attempted again
    
The goal here is to discuss solutions to these issues. 
1. Should `blobhandler.Accept` be made idempotent, such that it doesn't actually "store" the same blob twice?
2. How can the entire `Transfer` operation be made idempotent, such that it doesn't perform repeated transfers between nodes, and don't allocate blobs that have already been allocated?</div>

frrist:
- closes #215

One open question (cc @alanshaw) what is the expected behavior if the `Transfer` method never succeeds? Say, for example, it is never able to replicate the data from the source, due to the source not responding or something? At present if this case were to occur no receipt would be sent to the upload service - I assume that is not the desired outcome?


alanshaw:
> One open question (cc @alanshaw) what is the expected behavior if the `Transfer` method never succeeds? Say, for example, it is never able to replicate the data from the source, due to the source not responding or something? At present if this case were to occur no receipt would be sent to the upload service - I assume that is not the desired outcome?

Correct, piri should always send a receipt. It's a receipt where the outcome is an error.

`TransferReplicaBlobError` https://github.com/storacha/specs/blob/main/w3-replication.md#blob-replica-transfer-receipt-schema

pkg/service/storage/handlers/replica/transfer.go
```diff
@@ -223,13 +391,74 @@ func Transfer(ctx context.Context, service TransferService, request *TransferReq
...
+
+	// Create agent message with the failure receipt
+	agentMsg, err := message.Build([]invocation.Invocation{}, []receipt.AnyReceipt{rcpt})
+	if err != nil {
+		return fmt.Errorf("building failure message: %w", err)
+	}
+
+	// Send the failure receipt to the upload service
+	if err := sendMessageToUploadService(ctx, service, agentMsg); err != nil {
+		return fmt.Errorf("sending failure receipt: %w", err)
```
frrist, line 461: @alanshaw I am uncertain of this `TransferReplicaBlobError` type implementation. I _think_ it might require a change in go-libstoracha, but I am not sure. wdyt?

pkg/service/storage/handlers/replica/transfer.go
```diff
@@ -181,35 +283,123 @@ func Transfer(ctx context.Context, service TransferService, request *TransferReq
...
+// sendMessageToUploadService sends the message containing invocations and receipts to the upload service
+func sendMessageToUploadService(ctx context.Context, service TransferService, rcpt receipt.AnyReceipt) error {
+	concludeInv, err := ucan.Conclude.Invoke(
+		service.ID(),
+		service.UploadConnection().ID().DID(),
+		service.ID().DID().String(),
+		ucan.ConcludeCaveats{
+			Receipt: rcpt.Root().Link(),
+		},
+		delegation.WithNoExpiration(),
```
alanshaw, line 386: We don't need that to be valid forever (will default to 30s from now).

```suggestion
```

pkg/service/storage/handlers/replica/transfer.go
```diff
@@ -181,35 +283,123 @@ func Transfer(ctx context.Context, service TransferService, request *TransferReq
...
+	for rcptBlk, err := range rcpt.Blocks() {
+		if err != nil {
+			return fmt.Errorf("iterating receipt blocks: %w", err)
+		}
+		if err := concludeInv.Attach(rcptBlk); err != nil {
+			return fmt.Errorf("attaching receipt block: %w", err)
+		}
+	}
+
+	msg, err := message.Build([]invocation.Invocation{concludeInv}, []receipt.AnyReceipt{rcpt})
```
alanshaw, line 402: Why do you not call `client.Execute()` here?

```go
client.Execute(ctx, []invocation.Invocation{concludeInv}, service.UploadConnection())
```

alanshaw (approved):
LGTM

pkg/service/storage/handlers/replica/transfer.go
frrist: Primarily lack of familiarity with how I should use it...though it's canonical way to do this, so I'll switch over to it.


But for context, and to help me understand better:
```go
type ExecutionResponse interface {
	// Blocks returns an iterator of all the IPLD blocks that are included in
	// the response.
	Blocks() iter.Seq2[block.Block, error]
	// Get returns a link to a receipt, given an invocation link.
	Get(inv ucan.Link) (ucan.Link, bool)
}

func Execute(ctx context.Context, invocations []invocation.Invocation, conn Connection) (ExecutionResponse, error) {
	input, err := message.Build(invocations, nil)
	if err != nil {
		return nil, fmt.Errorf("building message: %w", err)
	}

	req, err := conn.Codec().Encode(input)
	if err != nil {
		return nil, fmt.Errorf("encoding message: %w", err)
	}

	res, err := conn.Channel().Request(ctx, req)
	if err != nil {
		return nil, fmt.Errorf("sending message: %w", err)
	}

	output, err := conn.Codec().Decode(res)
	if err != nil {
		return nil, fmt.Errorf("decoding message: %w", err)
	}

	return ExecutionResponse(output), nil
}
```

It looks like:
```go
res, err := conn.Channel().Request(ctx, req)
if err != nil {
	return nil, fmt.Errorf("sending message: %w", err)
}
```
will return an error with no response if the status code != 200. Which is basically what we want here, I think.

What should I do with the `ExecutionResponse` it returns for the success case? 

Other examples I see create a receipt reader, is this the [schema](https://github.com/storacha/specs/blob/eaf1bad1630dd83b2402828b26c9be4cad63cd0a/w3-ucan.md?plain=1#L234) I should expect back? Or would it be simpler to ignore the response for this case?

frrist: I'm not sure what I am doing wrong here, see f41569a8f67c015895340b11b89eb66bd9460dab for this change, though it breaks the tests, which may also be wrong?

PR #238, opened 2025-09-12
feat: UCAN authorized retrievals

alanshaw:
This PR adds UCAN authorized retrievals via `space/content/retrieve` invocations, as specified [here](https://github.com/storacha/specs/blob/main/w3-retrieval.md).

It adds a go-ucanto retrieval server, and with a handler for `space/content/retrieve`. The server will serve the blob from the blobstore provided it was allocated in the space referenced by the invocation.

In PDP mode, blobs are stored in a different store (or in a completely different process), keyed by piece hash so we use an adapter to allow the blobs keyed by piece hash to be accessed via user hash.

BREAKING CHANGE: allocations are now keyed by blob digest and space instead of blob digest and invocation CID.


pkg/pdp/store/adapter/blobgetter.go
```diff
@@ -0,0 +1,66 @@
...
+
+// BlobGetterAdapter adapts a PDP piece finder and piece reader into a
+// [blobstore.BlobGetter]
+type BlobGetterAdapter struct {
+	pieceFinder piecefinder.PieceFinder
+	pieceReader piecereader.PieceReader
+	blobSizer   BlobSizer
+}
+
+func (psa *BlobGetterAdapter) Get(ctx context.Context, digest multihash.Multihash, opts ...blobstore.GetOption) (blobstore.Object, error) {
```
volmedo, line 43:
```suggestion
func (bga *BlobGetterAdapter) Get(ctx context.Context, digest multihash.Multihash, opts ...blobstore.GetOption) (blobstore.Object, error) {
```

pkg/service/retrieval/ucan/space_content_retrieve.go
```diff
@@ -0,0 +1,108 @@
...
+						rangeNotSatisfiableErr := content.NewRangeNotSatisfiableError(fmt.Sprintf("range not satisfiable: %d-%d", start, end))
+						res := result.Error[content.RetrieveCaveats, failure.IPLDBuilderFailure](rangeNotSatisfiableErr)
+						resp := retrieval.NewResponse(http.StatusRequestedRangeNotSatisfiable, nil, nil)
+						return res, nil, resp, nil
+					}
+					log.Errorw("getting blob", "error", err)
+					return nil, nil, retrieval.Response{}, fmt.Errorf("getting blob: %w", err)
+				}
+
+				res := result.Ok[content.RetrieveCaveats, failure.IPLDBuilderFailure](content.RetrieveCaveats{})
```
volmedo, line 88: shouldn't this be ?
```suggestion
				res := result.Ok[content.RetrieveOk, failure.IPLDBuilderFailure](content.RetrieveOk{})
```

pkg/service/retrieval/ucan_fx_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/storacha/piri/pkg/fx/app"
+	piritestutil "github.com/storacha/piri/pkg/internal/testutil"
+	"github.com/storacha/piri/pkg/service/storage"
+	"github.com/storacha/piri/pkg/store/allocationstore/allocation"
+)
+
+func TestFXSpaceContentRetrieve(t *testing.T) {
+	var svc storage.Service
+
+	appConfig := piritestutil.NewTestConfig(t, piritestutil.WithSigner(testutil.Alice))
```
volmedo, line 32: I think using `testutil.Service` instead of `testutil.Alice` would avoid some confusion

pkg/store/allocationstore/util.go
```diff
@@ -0,0 +1,34 @@
+package allocationstore
+
+import (
+	"context"
+	"fmt"
+
+	multihash "github.com/multiformats/go-multihash"
```
volmedo, line 7: this alias is redundant. Maybe you meant to alias it to `mh`?

pkg/store/blobstore/interface.go
```diff
@@ -19,53 +19,63 @@ var ErrTooLarge = errors.New("payload too large")
 // ErrTooSmall is returned when the data being written is smaller than expected.
 var ErrTooSmall = errors.New("payload too small")
 
+// ErrRangeNotSatisfiable is returned when the byte range option falls outside
+// of the total size of the blob.
+var ErrRangeNotSatisfiable = errors.New("range not satisfiable")
+
 // GetOption is an option configuring byte retrieval from a blobstore.
-type GetOption func(cfg *options) error
+type GetOption func(cfg *GetOptions) error
```
volmedo, line 27: does `GetOptions` need to be exported?

pkg/store/blobstore/todo_datastore.go
volmedo, line 1: what is `TODO_DsBlobstore` used for?

volmedo (approved):

pkg/store/blobstore/todo_datastore.go
frrist: Context here: https://github.com/storacha/piri/pull/238#discussion_r2373123797

pkg/service/retrieval/ucan_fx_test.go
alanshaw: Yes I agree, but I did it this way because the storage tests do the same 😩

pkg/service/retrieval/ucan/space_content_retrieve.go
alanshaw: Yes, FML I just found that myself.

PR #240, opened 2025-09-15
Frrist/update 2

frrist:

PR #244, opened 2025-09-17
Fast follows to piri auto update

frrist:
A follow on to https://github.com/storacha/piri/pull/240/

This change does two things: 
It ensures when a rollback happens, we delete the binary causing the failure and re-install it.
Additionally, in #240 systemd weren't versioned, this will cause issues during updates when service requirements changed. i.e. Service files could get out of sync with binary versions. Maybe we want a longer shutdown timeout, etc.
So now we:
  - Versions systemd files in `/opt/piri/systemd/vX.Y.Z/` alongside binaries (same pattern)
  - Uses symlinks: `/opt/piri/systemd/current` -> `/opt/piri/systemd/vX.Y.Z`
  - Service files in `/etc/systemd/system/` symlink through current version
  - Updates both binary and systemd symlinks atomically during auto-update

### DIFF
--- a/cmd/cli/setup/config.go
+++ b/cmd/cli/setup/config.go
@@ -8,33 +8,35 @@ import (
 
 // PathConfig holds all path-related configuration
 type PathConfig struct {
-	OptDir           string        // Base directory for piri installation
-	BinaryBaseDir    string        // Base directory for all versioned binaries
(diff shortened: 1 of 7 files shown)


alanshaw (approved):
LGTM

Issue #174, opened 2025-08-06
Send "pay me" invocation in Piri to billing server


hannahhoward:
# What

After successfully serving a retrieval made with space/content/retrieve, the node should respond send a "pay me" invocation to the billing server

PR #258, opened 2025-09-25
feat: add receipts to the EgressTrackingService on retrieval

volmedo:
Ref: #174 

Depends on: #245 and #246 

Attach a ReceiptLogger to the `space/content/retrieve` handler to collect retrieval receipts and add them to the EgressTrackingService to be stored, batched and sent in `space/egress/track` invocations.

### DIFF
--- a/pkg/fx/retrieval/ucan/handlers/provider.go
+++ b/pkg/fx/retrieval/ucan/handlers/provider.go
@@ -47,36 +47,31 @@ func withErrorHandler() ucanretrieval.Option {
 }
 
 func withReceiptLogger(ets *egresstracking.EgressTrackingService) ucanretrieval.Option {
-	return ucanretrieval.WithReceiptLogger(func(_ context.Context, rcpt receipt.AnyReceipt, inv invocation.Invocation) {
+	return ucanretrieval.WithReceiptLogger(func(_ context.Context, rcpt receipt.AnyReceipt, inv invocation.Invocation) error {
 		// Egress Tracking is optional, the service will be nil if it is disabled
 		if ets == nil {
-			return
+			return nil
 		}
 
-		// Collect the receipt in a goroutine to avoid blocking the handler
-		go func() {
-			// Make sure the receipt is self-contained, i.e. it also has invocation blocks
-			fullRcpt, err := rcpt.Clone()
-			if err != nil {
-				log.Errorw("cloning receipt", "error", err)
-				return
-			}
+		// Make sure the receipt is self-contained, i.e. it also has invocation blocks
+		fullRcpt, err := rcpt.Clone()
+		if err != nil {
+			return err
+		}
 
-			if err := fullRcpt.AttachInvocation(inv); err != nil {
-				log.Errorw("attaching invocation to receipt", "error", err)
-				return
-			}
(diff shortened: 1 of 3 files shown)


pkg/fx/retrieval/ucan/handlers/provider.go
```diff
@@ -1,16 +1,82 @@
...
+		l := log.With("error", err.Error())
+		if s := err.Stack(); s != "" {
+			l = l.With("stack", s)
+		}
+		l.Error("ucan retrieval handler execution error")
+	})
+}
+
+func withReceiptLogger(ets *egresstracking.EgressTrackingService) ucanretrieval.Option {
+	return ucanretrieval.WithReceiptLogger(func(ctx context.Context, rcpt receipt.AnyReceipt, inv invocation.Invocation) {
```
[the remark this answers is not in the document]
volmedo: oh, that is the result, which is part of the receipt. But the receipt contains some other stuff required for the UCAN flow (see https://github.com/storacha/go-ucanto/blob/main/core/receipt/datamodel/receipt.ipldsch).

The handler handles the request and produces a result, and the ucanto server takes care of issuing a proper receipt by adding the other bits, do the signing, etc. If we want to get the receipt as it is sent to the client, we need to hijack the server process to fetch it, and this is what I did here by letting the user pass a callback that the server invokes right before sending the receipt back to the client.

pkg/fx/retrieval/ucan/handlers/provider.go
```diff
@@ -1,16 +1,82 @@
...
+			if err != nil {
+				log.Errorw("rebinding receipt", "error", err)
+				return
+			}
+
+			if err := ets.AddReceipt(context.Background(), retrievalRcpt); err != nil {
+				log.Errorw("adding receipt to egress tracking service", "error", err)
+				return
+			}
+		}()
```
frrist, line 80: I liked this better when we were passing the context from the caller here 😄 can we add it back?

volmedo: I agree with your concerns. A bounded pool or any other rate limiting mechanism would be the way to ensure there is some back-pressure on clients, but happy to remove the goroutine in the meantime.

About making the request fail if adding the receipt fails, there is a problem with the current implementation. `AddReceipt` enqueues a task in a jobqueue if there is a new batch to send, so this is happening asynchronously. If we think the client should not get its content delivered if recording the receipt fails (which is a fair point), then we shouldn't use the jobqueue and just implement a synchronous version of the task.

I created https://github.com/storacha/piri/issues/263 to help keeping this in mind.

frrist: For now lets remove the go routine.

As far as receipt recording failures:
- If the node is unable to `Append` a receipt to the store, due to lack of space, failure to rotate, etc. that's when I'd want to abort serving the data of the request. Since the node is unable to record a record of the content it served for payment. 
- Failing to report a batch to the egress service is also not ideal, but I view as a less critical error since a record of the content served will still exist on the filesystem, and manual intervention could be taken to "re-report" it for payment.

volmedo:
merging this into #246 before everything gets merged into #245. Happy to continue the discussion there.

PR #260, opened 2025-09-25
fix(aws): return store.ErrNotFound

hannahhoward:
# Goals

When a user performs an incomplete upload, where the blob is allocated but not received in full (i.e. http/put fails) -- they currently won't be able to upload again later, because we incorrectly see a blob store read error when the error is actually that the blob is not found when using the AWS blob store

# Implementation

- check for AWS no such key error and convert to ErrNotFound

### DIFF
--- a/pkg/aws/s3blobstore.go
+++ b/pkg/aws/s3blobstore.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


pkg/aws/s3blobstore.go
```diff
@@ -154,6 +157,11 @@ func (s *S3BlobStore) Get(ctx context.Context, digest multihash.Multihash, opts
 		Range:  rangeParam,
 	})
 	if err != nil {
+		var noSuchKeyError *types.NoSuchKey
+		// wrap in error recognizable as a not found error for Store interface consumers
+		if errors.As(err, &noSuchKeyError) {
+			return nil, store.NewErrNotFound(err)
```
alanshaw, line 163: Sorry needs more change - this *i think*:

```suggestion
			return nil, store.ErrNotFound
```

alanshaw (changes requested):

alanshaw (approved):

PR #264, opened 2025-10-01
feat: check consolidated records and clean receipt batches

volmedo:
Ref: #174 

Poll the etracker service periodically to fetch `space/egress/consolidate` receipts. This requires storing the consolidate invocation link sent in the `space/egress/track` receipt as an effect. It will be used later to fetch the corresponding receipt from the etracker receipts endpoint.

The original track invocation is also stored so that it can be used when checking the consolidate receipt (validation of the receipt still TODO).

Both the track invocation and the consolidate invocation link are stored in a new, datastore-based `ConsolidationStore`, keyed by the batch CID under two different namespaces.

Once consolidation has happened and the receipt has been confirmed to be correct, the batch can be removed from the journal.


pkg/store/consolidationstore/store.go
```diff
@@ -0,0 +1,150 @@
...
+// consolidate invocation CIDs, indexed by batch CID.
+// When a batch is tracked via space/egress/track, the receipt contains an effect
+// with a space/egress/consolidate invocation. We store both the track invocation
+// and the consolidate invocation CID for later retrieval.
+type Store interface {
+	// Put stores an egress/track invocation and consolidate invocation CID indexed by batch CID.
+	Put(ctx context.Context, batchCID cid.Cid, trackInv invocation.Invocation, consolidateInvCID cid.Cid) error
+
+	// GetTrackInvocation retrieves the egress/track invocation for a given batch CID.
+	GetTrackInvocation(ctx context.Context, batchCID cid.Cid) (invocation.Invocation, error)
```
frrist, line 32: I believe this method is unused, are there plans for it in the future?

pkg/store/consolidationstore/store.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return nil
+}
+
+func (cs *consolidationStore) GetTrackInvocation(ctx context.Context, batchCID cid.Cid) (invocation.Invocation, error) {
+	key := datastore.NewKey(batchCID.String())
+
+	data, err := cs.trackInvocationsDS.Get(ctx, key)
+	if err != nil {
+		if err == datastore.ErrNotFound {
```
frrist, line 73: prefer `error.Is`

pkg/store/consolidationstore/store.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return inv, nil
+}
+
+func (cs *consolidationStore) GetConsolidateInvocationCID(ctx context.Context, batchCID cid.Cid) (cid.Cid, error) {
+	key := datastore.NewKey(batchCID.String())
+
+	data, err := cs.consolidateInvCIDsDS.Get(ctx, key)
+	if err != nil {
+		if err == datastore.ErrNotFound {
```
frrist, line 109: ditto

pkg/store/consolidationstore/store.go
```diff
@@ -0,0 +1,150 @@
...
+	trackKey := datastore.NewKey(batchCID.String())
+	if err := cs.trackInvocationsDS.Put(ctx, trackKey, b); err != nil {
+		return fmt.Errorf("writing track invocation to datastore: %w", err)
+	}
+
+	// Store consolidate invocation CID
+	consolidateKey := datastore.NewKey(batchCID.String())
+	if err := cs.consolidateInvCIDsDS.Put(ctx, consolidateKey, consolidateInvCID.Bytes()); err != nil {
+		return fmt.Errorf("writing consolidate CID to datastore: %w", err)
+	}
```
frrist, line 63: Can use the same key as is done in `Delete` below

pkg/service/egresstracking/service.go
```diff
@@ -133,5 +149,151 @@ func (s *EgressTrackingService) egressTrack(ctx context.Context, batchCID cid.Ci
...
+		if err := s.checkAndRemoveConsolidatedBatch(ctx, batchCID); err != nil {
+			log.Errorf("error checking batch %s: %v", batchCID, err)
+			// Continue with other batches even if one fails
+		}
+	}
+
+	return nil
+}
+
+var ErrNotFound = errors.New("receipt not found")
```
frrist, line 249: unused?

pkg/service/egresstracking/service.go
```diff
@@ -133,5 +149,151 @@ func (s *EgressTrackingService) egressTrack(ctx context.Context, batchCID cid.Ci
...
+	log.Debugf("batch %s removed from store", batchCID.String())
+
+	return nil
+}
+
+func (s *EgressTrackingService) validateConsolidateReceipt(receipt receipt.AnyReceipt) error {
+	// TODO: Validate the receipt. This will include checking the receipt matches the original track invocation
+	// and confirming that the consolidated amount of bytes matches our records.
 	return nil
 }
```
frrist, line 299: Happy for this implementation to arrive in this PR too.

frrist (commented):
Would like to see the impl for `validateConsolidateReceipt` land here too, suspect it will make review easier.

pkg/store/consolidationstore/store.go
volmedo: yes, the idea is that the original track invocation will be needed to confirm the validity of the consolidate receipt for proper accounting in a trustless manner.

volmedo:
I'll implement `validateConsolidateReceipt` once I implement the egress tracker side of consolidation. It will likely involve updating the spec and I think it's important to get it right, and I want to deliver value incrementally. The fact that validation logic is defined in its own method should help isolating it from the rest of the code to not be an impediment for review. But happy to provide you with any details about validation that would make review easier.

The current changes set everything up and "work" in the sense that the full egress tracking flow is completed and can be tested. The fact that `validateConsolidateReceipt` is not doing much now just means that the nodes trust the accounting performed by the egress tracker. I created #265 so that proper validation is implemented.

Would you agree to merge this as is for now so that most of the flow can be tested?

pkg/client/receipts/client.go
```diff
@@ -0,0 +1,93 @@
+package receipts
```
frrist, line 1: Feels like this package should live in `pkg/service/egresstracking` since it's only used there.

pkg/service/egresstracking/service.go
```diff
@@ -128,5 +144,149 @@ func (s *EgressTrackingService) egressTrack(ctx context.Context, batchCID cid.Ci
...
+
+	log.Debugf("checking %d batches for consolidation", len(batchCIDs))
+
+	// Check each batch for consolidation
+	for _, batchCID := range batchCIDs {
+		if err := s.checkAndRemoveConsolidatedBatch(ctx, batchCID); err != nil {
+			log.Errorf("error checking batch %s: %v", batchCID, err)
+			// Continue with other batches even if one fails
+		}
+	}
```
frrist, line 238: Noticing that we have an opportunity to perform this operation in parallel (https://pkg.go.dev/golang.org/x/sync/errgroup is handy for this sort of thing), but also get that we don't (know if we) need that yet. Considering adding a comment here for our future selves.

pkg/service/egresstracking/service.go
```diff
@@ -128,5 +144,149 @@ func (s *EgressTrackingService) egressTrack(ctx context.Context, batchCID cid.Ci
...
+
+	log.Debugf("consolidate receipt fetched for batch %s", batchCID.String())
+
+	// Fetch the original track invocation for the batch
+	trackInv, err := s.consolidationStore.GetTrackInvocation(ctx, batchCID)
+	if err != nil {
+		return fmt.Errorf("batch %s not found in consolidation store: %w", batchCID, err)
+	}
+
+	s.validateConsolidateReceipt(rcpt, trackInv)
```
frrist, line 270: Even though this method doesn't return an error yet, lets include the error check anyways so we don't forget when we complete the `validateConsolidateReceipt` implementation.

pkg/service/egresstracking/service.go
```diff
@@ -128,5 +144,149 @@ func (s *EgressTrackingService) egressTrack(ctx context.Context, batchCID cid.Ci
...
+	// Remove the batch from the store
+	log.Infof("consolidate receipt found for batch %s, removing from store", batchCID)
+	if err := s.journal.Remove(ctx, batchCID); err != nil {
+		return fmt.Errorf("removing consolidated batch: %w", err)
+	}
+
+	// Remove from consolidation store
+	if err := s.consolidationStore.Delete(ctx, batchCID); err != nil {
+		log.Warnf("failed to remove batch %s from consolidation store: %v", batchCID, err)
+	}
```
frrist, line 281: Why return an error for `journal.Remove` but only log for `consolidationStore.Delete`? Just curious.

pkg/service/egresstracking/service.go
```diff
@@ -128,5 +144,149 @@ func (s *EgressTrackingService) egressTrack(ctx context.Context, batchCID cid.Ci
 		return fmt.Errorf("invocation failed: %s", x.Message)
 	}
 
+	// Extract the consolidate invocation from the receipt's effects
+	effects := rcpt.Fx()
+	if effects != nil {
+		fork := effects.Fork()
```
frrist, line 150: What would it mean for `effects` to be `nil`? Indicative of a developer error in the egress tracker?

frrist (approved):
LGTM, might be worth waiting for Alan to look at the UCAN bits, up to you.

alanshaw (approved):

pkg/service/egresstracking/service.go
volmedo: it does actually return an error now, I just forgot handling it 😄

pkg/service/egresstracking/service.go
volmedo: failing to remove the batch from the journal means it will be checked again. OTOH, failing to remove from the consolidation store does no harm beyond the records being kept in the datastore. It's a subtle difference, happy to change the log into an error if you think that makes more sense.

PR #266, opened 2025-10-02
feat: implement provider registration

frrist:
# Summary

Adds support for registering PDP service providers with the ServiceProviderRegistry smart contract, this is a prerequist for interaction with the service contract.

**Please see the bottom of this PR description, I have open questions and provide more context on next steps**

# Changes

## Service Layer
  - pkg/pdp/service/provider_register.go: Implements RegisterProvider() method
    - Checks if provider is already registered using IsRegisteredProvider()
    - Returns existing provider info if already registered
    - Constructs and sends registration transaction with provider metadata (name, description, PDP offering)
    - Stores transaction hash in database for tracking

## HTTP API

  - Server (pkg/pdp/httpapi/server/register_provider.go):
    - POST /pdp/provider/register - Registers provider with optional name and description
    - Returns provider details including transaction hash, address, payee, ID, and status
  - Client (pkg/pdp/httpapi/client/client.go):
    - RegisterProvider() method for programmatic access
  - Types (pkg/pdp/httpapi/types.go):
    - RegisterProviderRequest - name and description fields
    - RegisterProviderResponse - comprehensive provider information

## CLI

  - cmd/cli/client/pdp/provider/: New provider command group
    - piri client pdp provider register --name <name> --description <desc>
    - Outputs JSON with provider registration results
    - Both flags are required

## Smart Contract Integration
  - Generate Bindings (pkg/pdp/smartcontracts/bindings/service_provider_registry.go): These are based on our fork of the contract in https://github.com/storacha/filecoin-services/blob/main/service_contracts/src/ServiceProviderRegistry.sol will try and have CI setup for this to make the abi generation less manual.

## Usage

**Register as a provider:**
```bash
  piri client pdp provider register \
    --name "My Provider" \
    --description "Reliable storage with 90 9's of uptime or whatever"
```
**Response**:
```json
  {
    "txHash": "0x...",
    "address": "0x...",
    "payee": "0x...",
    "id": 1,
    "isActive": true,
    "name": "My PDP Provider",
    "description": "Reliable storage with 90 9's of uptime or whatever"
  }
```

## Technical Notes
  - Registration requires 5 FIL fee (burned by contract)
    - We (cc @hannahhoward @alanshaw) may want to consider this fee: It serves as an economic spam prevention mechanism. The fee is burned by sending it to the `BURN_ACTOR` address, ofc this amount is configurable via contract modification, as is the location the fee is sent too.
  - Method is idempotent - calling multiple times returns existing registration
    - **caveat** here is that calling the method more than once in an epoch will result in wasted gas. We could put the work in now to prevent this, as my comment in code outlines, but I am being lazy for the time being, and overloading this method as both a registration method, and status method. 
  - Provider ID is assigned by the contract upon successful registration - its a counter that just goes up, similar to data set IDs

## Work to follow
Once a provider is registered, we as the owners of the service contract must approve the provider. Obv this isn't code we'll add to this repo, feedback welcome on where the "approval" mechanism should land. Note that "approvals" are sent to the warm storage contract address, not the registry. 

Getting into the details a bit, we can register a provider using the `cast` cli too (part of foundry) like this:
```bash
cast send $FILECOIN_WARM_STORAGE_SERVICE_ADDRESS \
    "addApprovedProvider(uint256)" \
    $PROVIDER_ID \
    --rpc-url $RPC_URL \
    --private-key $OWNER_PRIVATE_KEY
```
Where `RPC_URL` is a lotus node (e.g. Glif RPC) and `OWNER_PRIVATE_KEY` is the sk of the wallet we deployed the contract with, which I have shared out-of-band to this work.

### DIFF
--- a/pkg/pdp/service/provider_register.go
+++ b/pkg/pdp/service/provider_register.go
@@ -1,0 +1,161 @@
+package service
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/storacha/piri/pkg/pdp/types"
+	"gorm.io/gorm"
+
+	"github.com/storacha/piri/pkg/pdp/service/models"
+	"github.com/storacha/piri/pkg/pdp/smartcontracts"
+	"github.com/storacha/piri/pkg/pdp/smartcontracts/bindings"
+)
+
+func (p *PDPService) RegisterProvider(ctx context.Context, params types.RegisterProviderParams) (types.RegisterProviderResults, error) {
+	// TODO(forrest): remove this once confident in the code
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("Recovered in RegisterProvider", r)
+		}
+	}()
+	bindCtx := &bind.CallOpts{Context: ctx}
+	registry, err := bindings.NewServiceProviderRegistry(smartcontracts.Addresses().ProviderRegistry, p.contractBackend)
+	if err != nil {
+		return types.RegisterProviderResults{}, fmt.Errorf("failed to create service registry binding: %w", err)
+	}
+
+	isRegistered, err := registry.IsRegisteredProvider(bindCtx, p.address)
+	if err != nil {
+		return types.RegisterProviderResults{}, fmt.Errorf("failed to check if service provider is registered: %w", err)
+	}
+
+	if isRegistered {
+		// TODO we can move this to a separate method for query provider, doing this here because its easy and I lazy.
+		providerInfoView, err := registry.GetProviderByAddress(bindCtx, p.address)
+		if err != nil {
+			return types.RegisterProviderResults{}, fmt.Errorf("failed to get provider by address for registered provider: %w", err)
+		}
+		log.Errorf("service provider %s is already registered with address %s", providerInfoView.ProviderId, p.address)
+		return types.RegisterProviderResults{
+			Address:     providerInfoView.Info.ServiceProvider,
+			Payee:       providerInfoView.Info.Payee,
+			ID:          providerInfoView.ProviderId.Uint64(),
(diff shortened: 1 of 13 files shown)


pkg/pdp/service/provider_register.go
```diff
@@ -0,0 +1,161 @@
...
+		IpniPiece:                  false,
+		IpniIpfs:                   false,
+		StoragePricePerTibPerMonth: big.NewInt(2),
+		MinProvingPeriodInEpochs:   big.NewInt(30),
+		Location:                   "earth",
+		PaymentTokenAddress:        p.address,
+	})
+	if err != nil {
+		return types.RegisterProviderResults{}, fmt.Errorf("failed to encode product data: %w", err)
+	}
```
frrist, line 114: Based on the contract code, the only field that actually matters is the payee: https://github.com/storacha/filecoin-services/blob/main/service_contracts/src/FilecoinWarmStorageService.sol#L492, which isn't in this struct, but defined below on line 116 as `p.address` - the address of the operator running the node.

I don't know if we want to let providers populate this with their own preferences (that don't actually matter), or hard code it with (less sarcastic) values than I have used here.

pkg/pdp/service/provider_register.go
```diff
@@ -0,0 +1,161 @@
...
+		  capabilities
+		  2. The actual service contract (FilecoinWarmStorageService) enforces its
+		  own standardized terms
+		  3. Clients might use PDPOffering data to discover providers, but the
+		  actual service operates on fixed terms
+	*/
+	productData, err := registry.EncodePDPOffering(bindCtx, bindings.ServiceProviderRegistryStoragePDPOffering{
+		// so I don't think any of these fields matter, see above comment
+		// TODO validate this assumption, I don't think it's used, but need to verify
+		ServiceURL:                 "http://example.com",
```
alanshaw, line 102: Should we at least put `https://storacha.network` here incase this is displayed somewhere?

alanshaw (approved):
LGTM

PR #273, opened 2025-10-08
feat: implement `access/grant` capability

alanshaw:
This PR implements an `access/grant` capability handler allowing agents to request `blob/retrieve` delegations, which in turn allows them to retrieve blobs in order to complete tasks.

Piri will now issue `blob/retrieve` delegations for these tasks:

1. A storage node that receives a `blob/replica/allocate` and needs to transfer the blob from another node.
1. An indexer that receives an `assert/index` invocation and needs to retrieve the index so that it can cache it's contents and add them to IPNI.

### DIFF
--- a/pkg/service/storage/ucan_fx_test.go
+++ b/pkg/service/storage/ucan_fx_test.go
@@ -38,7 +38,9 @@ import (
 	"github.com/storacha/go-ucanto/did"
 	ucanserver "github.com/storacha/go-ucanto/server"
 	"github.com/storacha/go-ucanto/ucan"
+	"github.com/storacha/go-ucanto/validator"
 	appconfig "github.com/storacha/piri/pkg/config/app"
+	"github.com/storacha/piri/pkg/principalresolver"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/fx"
 	"go.uber.org/fx/fxtest"
@@ -437,7 +439,7 @@ func TestFXReplicaAllocateTransfer(t *testing.T) {
 
 			appConfig := piritestutil.NewTestConfig(t,
 				piritestutil.WithSigner(testutil.Alice),
-				piritestutil.WithUploadServiceConfig(testutil.Service.DID(), uploadServiceURL),
+				piritestutil.WithUploadServiceConfig(testutil.WebService.DID(), uploadServiceURL),
 			)
 
 			testApp := fxtest.New(t,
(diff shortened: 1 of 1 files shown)


pkg/fx/principalresolver/provider.go
```diff
@@ -27,10 +28,14 @@ var Module = fx.Module("principalresolver",
...
-		services = append(services, idxSvc.ID().DID())
+		if strings.HasPrefix(idxSvc.ID().DID().String(), "did:web:") {
+			services = append(services, idxSvc.ID().DID())
+		}
 	}
 	if uplSvc := cfg.UCANService.Services.Upload.Connection; uplSvc != nil {
-		services = append(services, uplSvc.ID().DID())
+		if strings.HasPrefix(uplSvc.ID().DID().String(), "did:web:") {
+			services = append(services, uplSvc.ID().DID())
+		}
```
frrist: only adding web dids to the resolver since they are the only dids that can be resolved, yeah?

pkg/service/storage/ucan/access_grant.go
```diff
@@ -0,0 +1,229 @@
...
+	// invocation has been determined to be invalid is some way. See the error
+	// message for details.
+	InvalidCauseErrorName = "InvalidCause"
+	// UnauthorizedCauseErrorName is the name given to an error where the cause
+	// invocation failed UCAN validation.
+	UnauthorizedCauseErrorName = "UnauthorizedCause"
+)
+
+// validity is the time a granted delegation is valid for.
+const validity = time.Hour
```
frrist: Might be nice to plumb this into the configuration, though in practice seems fine. Maybe a TODO? Your call.

pkg/service/storage/service.go
```diff
@@ -293,6 +300,21 @@ func New(opts ...Option) (*StorageService, error) {
...
+			id.Verifier(),
+			validator.IsSelfIssued,
+			func(context.Context, validator.Authorization[any]) validator.Revoked {
+				return nil
+			},
+			validator.ProofUnavailable,
+			edverifier.Parse,
+			validator.FailDIDKeyResolution,
+		)
+	}
```
frrist: What happens when this isn't configured? Seems like behavior undefined?

pkg/service/storage/ucan/access_grant.go
alanshaw: I think we should be in charge of this value - i.e. don't let it be configured...so that we can change it later to get a smal perf bump if we wish.

pkg/fx/principalresolver/provider.go
alanshaw: I did that ^^ and reverted the change.

PR #276, opened 2025-10-08
fix: concurrent IPNI advert publishing

alanshaw:
This PR brings in the new IPNI publisher library that has a store with a `Replace` method (see https://github.com/storacha/go-libstoracha/pull/56).

The replace method allows the head to be updated safely when the backing store is an S3 bucket that might be concurrently written to by multiple lambdas.

The replace method makes use of `If-Match` HTTP header in order to only write the value if the current value has the same `Etag`.

We also try really hard to write an advert - this PR adds retries to the advert publish call that allows the conditional write to fail up to 5 times.


pkg/aws/s3store_test.go
```diff
@@ -0,0 +1,98 @@
...
+	paws "github.com/storacha/piri/pkg/aws"
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/modules/minio"
+)
+
+func TestS3StoreReplace(t *testing.T) {
+	if os.Getenv("CI") != "" && runtime.GOOS != "linux" {
+		t.SkipNow()
+	}
```
frrist: I am getting failures on OSX locally when running this test, was the intention for this to only run when both CI is set and the OS is Linux?

PR #286, opened 2025-10-16
feat: authorized retrieval for replica transfer

alanshaw:
This PR impements authorized retrieval for replica transfer.

In order to authorize retrieval it first makes a `access/grant` invocation to the replication source requesting a `blob/retrieve` delegation. It attaches the `blob/replica/allocate` invocation as evidence to the source that the request is legitimate.

It then uses the delegation in a `blob/retrieve` invocation to the source, allowing retrieval of the data.


pkg/service/storage/ucan_test.go
frrist: I wonder if we should just delete this? ucan_fx_test.go has the same coverage, or it _should_

pkg/service/storage/ucan/replica_allocate.go
```diff
@@ -128,7 +128,7 @@ func ReplicaAllocate(storageService ReplicaAllocateService) server.Option {
 				if err := storageService.Replicator().Replicate(ctx, &replicahandler.TransferRequest{
 					Space:  cap.Nb().Space,
 					Blob:   cap.Nb().Blob,
-					Source: replicaAddress,
+					Source: replicahandler.TransferSource{ID: claim.Issuer(), URL: replicaAddress},
```
frrist: Confirming my understanding: Issuer of the location claim is the node that must have published the content.

pkg/service/storage/handlers/replica/transfer.go
```diff
@@ -229,30 +249,76 @@ func checkBlobExists(ctx context.Context, service TransferService, blob types.Bl
...
+		return nil, err
+	}
+
+	rcpt, err := rcptReader.Read(rcptLink, replicaExecResp.Blocks())
+	if err != nil {
+		return nil, fmt.Errorf("reading %s receipt: %w", blob.RetrieveAbility, err)
+	}
+
+	_, x := result.Unwrap(rcpt.Out())
+	if x != (blob.RetrieveError{}) {
```
frrist: I _think_ `errors.Is` is safer.

PR #297, opened 2025-10-22
chore: smaller egress tracking batches for testing

volmedo:
Set batch size to 10KiB in the default config so that nodes produce batches more frequently during testing.

This is a follow-up to #296. I'll rebase on main once that one's merged.


frrist:
Please merge this with https://github.com/storacha/piri/pull/296

PR #301, opened 2025-10-26
feat: point to the registrar service at its new location

volmedo:
The registrar service has been deployed to the standard `warm.storacha.network` domain.


alanshaw (approved):

alanshaw:
We verified this worked in the test we ran yesterday.

PR #304, opened 2025-10-27
feat: flatfs object store

alanshaw:
This is the FlatFS datastore from https://github.com/ipfs/go-ds-flatfs without the disk usage reporting, batching, querying or windows support.

It has been modified slightly to implement `objectstore.Store`. Mainly it supports streaming and byte range requests. It has also been changed to allow lowercase keys instead of uppercase keys.

Notably, keys are restricted to the characters `0-9a-z+-_=`. There are a couple of existing tests that are skipped for FlatFS because of this.

I also added a thin wrapper around it (`blobstore.ObjectBlobstore` - I know great name) which allows it to satisfy the `blobstore.Blobstore` interface. The wrapper simply converts the multihash digest keys into base32 (lower, no pad) encoded strings.

I have replaced the `PDPStore` with this backing implementation.

### DIFF
--- a/pkg/store/objectstore/flatfs/flatfs.go
+++ b/pkg/store/objectstore/flatfs/flatfs.go
@@ -113,8 +113,10 @@ func (m *opMap) Begin(name string) *opResult {
 		// the result
 		op.mu.RLock()
 		if op.success {
+			op.mu.RUnlock()
 			return nil
 		}
+		op.mu.RUnlock()
 
 		// if we are here, we will retry the operation
 	}


pkg/store/objectstore/flatfs/flatfs.go
```diff
@@ -0,0 +1,500 @@
+// Package flatfs is a Datastore implementation that stores all
```
frrist: Is this mostly a copy from an existing store? i.e. how closely do we need to review this change?

alanshaw: Worth reviewing the `Get` method, and also the `Put` method (albeit with lesser priority).

The material change here is that the signatures for these methods have changed. If they hadn't changed we'd just be depend directly on the `go-ds-flatfs` library.

The `Get` method is very different because instead of returning the bytes, it returns an intermediate object that allows accessing the bytes (as a reader) as well as the total file size. It also accepts and deals with byte ranges in options. The implementation is almost identical to what we have in [the fs blobstore](https://github.com/storacha/piri/blob/662b1d1fca68df411afbe7de5b9f574f7870f451/pkg/store/blobstore/fs.go#L81) so if you're familiar with that you'll feel right at home ;)

The `Put` method is almost the same, but instead of accepting `[]byte`, it accepts a size and a reader. If you follow this through the change essentially manifests as a switch from `file.Write(bytes)` to `io.Copy(file, reader)`.

pkg/store/objectstore/flatfs/flatfs.go
```diff
@@ -0,0 +1,500 @@
...
+	ErrShardingFileMissing = fmt.Errorf("%s file not found in datastore", SHARDING_FN)
+	ErrClosed              = errors.New("datastore closed")
+	ErrInvalidKey          = errors.New("key not supported by flatfs")
+)
+
+// Store implements [objectstore.Store].
+// Note this datastore cannot guarantee order of concurrent
+// write operations to the same key. See the explanation in
+// Put().
+type Store struct {
```
frrist: A `Has` method would be really nice to have, specifically for https://github.com/storacha/piri/issues/323

alanshaw: 🙏 please can we add if we really need it? I don't think there's a big overhead with using `Get` for this - it's not going to start reading the file until you call `Body()` on the returned object.

PR #307, opened 2025-10-28
fix:(pdp): calculate piece offset per root & correct aggregate pieceCIDV2 Size Calculation

frrist:
# Fix Global Offset Accumulation & Aggregate PieceCIDV2 Calculation

#### Root Cause:

The bug originated in https://github.com/storacha/piri/blob/ff0ea2d386ee1d3da166a6c88a1332a18fb0e125/pkg/pdp/service/roots_add.go#L307-L360 where the `globalOffset` variable was being calculated globally across all roots in a single `AddRoots` operation. When adding multiple roots in one transaction, the offsets would accumulate:

When adding serveral roots in a single transaction, this resulted in, as an example:

  - Root 1: subroots at offsets 0, 134217728, 268435456...
  - Root 2: subroots at offsets 402653184, 536870912... (continuing from Root 1) (wrong)
  - Root 3: subroots at offsets 671088640, 805306368... (continuing from Root 2) (wrong)

While not the cause of the bug, the addition of https://github.com/storacha/piri/pull/294 which aggregates multipule roots into a single `AddRoot` call lead to the `AddRoot` method being called with more than one root, logic the method had never exercised prior to the addition of #294.

## How It Manifested in Proving

The verifier contract's [`FindPieceIds`](https://github.com/storacha/piri/blob/ff0ea2d386ee1d3da166a6c88a1332a18fb0e125/pkg/pdp/tasks/prove_pdp.go#L274) method returns per-root offsets - each root's pieces start at offset 0 within that root. However, Piri submitted roots with global offsets from the inital `AddRoots` operations, as outlined above. This is the bug minifesting.

  When the PDP Prove task attempted to prove a challenge, using the above set of roots as an example:

1. `FindPieceIds` would return something like: {PieceId: 2, Offset: 0} (meaning Root 2, starting at its offset 0)
2. The database query would look for offset 0 in Root 2's subroots
3. But Root 2's subroots started at offset 402653184 (or similar) in the database
4. Result: "no subroot found" error, making challenges impossible to prove

The error manifested as, for example:` failed to prove piece 1 (dataSetId: 22, pieceId: 1, leafIndex: 7846337): no subroot found`

### The Fix

Fixed Offset Calculation (roots_add.go line 299):

```go
for _, addReq := range request {
  // Reset offset for each root so subroots start at 0 for each root
  var totalOffset uint64 = 0  // NOW INSIDE the loop - resets per root
  for i, subCID := range addReq.SubRoots {
    subInfo.SubrootOffset = totalOffset
    totalOffset += uint64(subInfo.PieceInfo.Size)
  }
}
```

#### Why and how this works
1. Offset Alignment: Each root's subroots now start at offset 0, matching what the smart contract expects
2. Correct Challenge Resolution: When `FindPieceIds` returns a per-root offset, it now correctly matches the offsets Piri is aware of in its database.

The proving logic in `prove_pdp.go` was already correct - it was expecting per-root offsets. The bug was entirely in how offsets were stored during the `AddRoots` operation.

This fix ensures that:
- Multiple roots can be added in a single transaction
- Each root maintains its own offset space starting from 0
- Challenges can be correctly proven across all roots in a proof set
- The database schema properly reflects the semantic structure of the data

## Fix Aggregate PieceCIDV2 Calculation
Ensure the unpadded size for aggregates are calculate as:
`sum(subpiece(1).paddedSize ... subpiece(n-1).paddedSize) + subpiece(n).rawSize`

Previously the size of the entire aggregated piece zero padded out to a power of 2 was used, which was incorrect and resulted in challenges from the verifier contract over ranges of data that did not exist.

----------

closes https://github.com/storacha/piri/issues/317

### DIFF
--- a/pkg/pdp/smartcontracts/verifier.go
+++ b/pkg/pdp/smartcontracts/verifier.go
@@ -9,10 +9,13 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
+	logging "github.com/ipfs/go-log/v2"
 	"golang.org/x/xerrors"
 
 	"github.com/storacha/filecoin-services/go/bindings"
 )
+
+var log = logging.Logger("smartcontracts")
 
 type Verifier interface {
 	GetChallengeFinality(ctx context.Context) (*big.Int, error)
@@ -81,7 +84,12 @@ func (v *verifierContract) GetDataSetStorageProvider(ctx context.Context, setId 
 }
 
 func (v *verifierContract) GetChallengeRange(ctx context.Context, setId *big.Int) (*big.Int, error) {
-	return v.verifier.GetChallengeRange(&bind.CallOpts{Context: ctx}, setId)
+	out, err := v.verifier.GetChallengeRange(&bind.CallOpts{Context: ctx}, setId)
+	if err != nil {
+		return nil, err
+	}
+	log.Infow("verifier challenge range", "setId", setId, "range", out.Uint64())
+	return out, nil
 }
 
 func (v *verifierContract) GetScheduledRemovals(ctx context.Context, setId *big.Int) ([]*big.Int, error) {
@@ -89,7 +97,21 @@ func (v *verifierContract) GetScheduledRemovals(ctx context.Context, setId *big.
 }
 
 func (v *verifierContract) FindPieceIds(ctx context.Context, setId *big.Int, leafIndexs []*big.Int) ([]bindings.IPDPTypesPieceIdAndOffset, error) {
-	return v.verifier.FindPieceIds(&bind.CallOpts{Context: ctx}, setId, leafIndexs)
+	out, err := v.verifier.FindPieceIds(&bind.CallOpts{Context: ctx}, setId, leafIndexs)
+	if err != nil {
+		return nil, err
+	}
+
+	// Format out for human-readable logging
+	pieceInfo := make([]map[string]interface{}, len(out))
+	for i, piece := range out {
+		pieceInfo[i] = map[string]interface{}{
+			"pieceId": piece.PieceId.String(),
+			"offset":  piece.Offset.String(),
+		}
+	}
+	log.Infow("FindPieceIds", "setId", setId, "leafIndexs", leafIndexs, "pieces", pieceInfo)
+	return out, nil
 }
 
 func (v *verifierContract) GetNextPieceId(ctx context.Context, setId *big.Int) (*big.Int, error) {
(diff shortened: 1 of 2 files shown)


pkg/pdp/service/roots_add.go
```diff
@@ -383,105 +367,27 @@ func (p *PDPService) AddRoots(ctx context.Context, id uint64, request []types.Ro
 			return common.Hash{}, fmt.Errorf("computing height and padding: %w", err)
 		}
 		// NB: defined here: https://github.com/FilOzone/pdp/blob/main/src/PDPVerifier.sol#L44
-		// as MAX_PIECE_SIZE_LOG2 = 50
```
frrist: Removed hard coded values, and rely on the value the contract abi expects instead. removes a "magic" number.

pkg/pdp/service/roots_add.go
```diff
@@ -383,105 +367,27 @@ func (p *PDPService) AddRoots(ctx context.Context, id uint64, request []types.Ro
...
-		var totalUnpaddedSize uint64 = 0
 		var prevSubrootSize = subrootInfoMap[addRootReq.SubRoots[0]].PieceInfo.Size
 		for i, subrootEntry := range addRootReq.SubRoots {
 			subrootInfo := subrootInfoMap[subrootEntry]
 			if subrootInfo.PieceInfo.Size > prevSubrootSize {
+				// implies a bad request
 				return common.Hash{}, fmt.Errorf("subroots must be in descending order of size, root %d %s is larger than prev subroot %s", i, subrootEntry, addRootReq.SubRoots[i-1])
 			}
 
-			prevSubrootSize = subrootInfo.PieceInfo.Size
```
alanshaw: We should update this no? Otherwise we're always comparing against the first piece not the previous piece.

pkg/pdp/smartcontracts/verifier.go
```diff
@@ -55,6 +56,10 @@ func NewVerifierContract(backend bind.ContractBackend) (Verifier, error) {
 	}, nil
 }
 
+func (v *verifierContract) MaxPieceSizeLog2(ctx context.Context) (*big.Int, error) {
+	return v.verifier.MAXPIECESIZELOG2(&bind.CallOpts{Context: ctx})
```
volmedo: I don't know what this does, but it looks like it could be an expensive call? If that's the case, the value of `MaxPiecesSizeLog2` could be cached/memoized to avoid doing it every time. I'm assuming it's not a value that will change that often.

volmedo (commented):
I don't understand most of what's going on here, but left some comments in an attempt to be helpful. Feel free to disregard if not.

pkg/pdp/service/roots_add.go
frrist: ahh great catch, yeah this is important for validation, will fix.

pkg/pdp/smartcontracts/verifier.go
frrist: https://github.com/storacha/piri/issues/308

PR #318, opened 2025-10-30
fix: increase allowed transfer time

alanshaw:
This PR increases the transfer task expiration from 30 seconds to 1 hour. It allows for queuing delays and retries. This time is effectively the time we give ourselves to transfer the blob and send a `ucan/conclude` invocation to the upload service.

Aims to resolve:

```
'Unauthorized: Claim {"can":"blob/replica/transfer"} is not authorized\n' +
          '  - Proof bafyreig3wen5tsemfffi6u4tfou4y7v2mffajuyujgc53vpvoykbzdf6pe has expired on Thu Oct 30 2025 16:17:30 GMT+0000 (Coordinated Universal Time)\n'
```

...which is the upload service validating the transfer invocation in the `ucan/conclude` invocation.

### DIFF
--- a/pkg/service/storage/ucan/replica_allocate.go
+++ b/pkg/service/storage/ucan/replica_allocate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"time"
 
 	"github.com/storacha/go-libstoracha/capabilities/assert"
 	"github.com/storacha/go-libstoracha/capabilities/blob/replica"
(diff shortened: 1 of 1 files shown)


pkg/service/storage/ucan/replica_allocate.go
```diff
@@ -101,6 +104,7 @@ func ReplicaAllocate(storageService ReplicaAllocateService) server.Option {
 						Site:  cap.Nb().Site,
 						Cause: inv.Link(),
 					},
+					delegation.WithExpiration(ucan.Now()+transferTimeout),
```
volmedo: nit: what about making `const transferTimeout = 1 * time.Hour` and then

```suggestion
					delegation.WithExpiration(ucan.UTCUnixTimestamp(time.Now().Add(transferTimeout).Unix())),
```

volmedo (approved):

frrist (approved):

PR #322, opened 2025-10-31
tests: remove FX deps logs for test setup

frrist:
These have been driving me nuts and make CI logs too verbose.


alanshaw (approved):
Thank you!

Issue #323, opened 2025-11-03
Key blobs by client hash


alanshaw:
In Piri we store blobs using the Piece CID as the key.

This poses a number of problems:

1. We don't know the piece CID ahead of receiving the data, so we have to calculate the piece CID during upload, which is significantly slower than calculating alternative hashes like SHA2-256. This places a bottleneck on the effective upload speed of a node.
2. We don't know the piece CID ahead of receiving the data, yet we use the piece CID as the key in the data store, so we must buffer ALL the data into memory (#313), calculate the piece CID and then use it for the key.
3. The piece CID _must_ be calculated in-process at the time of the upload in order to add to the store, and yet the piece is aggregated asynchronously with other pieces and eventually added to PDP.
4. For retrieval, we have to lookup the piece CID that corresponds to the client supplied hash in order to retrieve the blob from the store. This places a small but additional overhead on the RTT for retrieving a blob. Note that we'll likely do this many times over the lifetime of the data, whereas we only need to use the piece CID once when adding to PDP (and once again when removing).

## Proposal: key blobs by client hash

1. Store blobs keyed by client supplied hash.
2. Calculate piece CID async _after_ upload completes and then start PDP aggregation.
3. Continue to maintain mapping of client hash to piece CID, but do not require using it in the retrieval hot path.

If users want to use a piece CID as their retrieval hash they can!

PR #330, opened 2025-11-06
Implement async commp and propoer piece resolution

frrist:
closes https://github.com/storacha/piri/issues/323

### DIFF
--- a/pkg/pdp/piece/commp.go
+++ b/pkg/pdp/piece/commp.go
@@ -11,11 +11,12 @@ import (
 	"github.com/multiformats/go-multihash"
 	captypes "github.com/storacha/go-libstoracha/capabilities/types"
 	"github.com/storacha/go-libstoracha/piece/piece"
-	"github.com/storacha/piri/pkg/pdp/aggregator"
-	"github.com/storacha/piri/pkg/pdp/aggregator/jobqueue"
-	"github.com/storacha/piri/pkg/pdp/aggregator/jobqueue/serializer"
-	"github.com/storacha/piri/pkg/pdp/types"
 	"go.uber.org/fx"
+
+	"github.com/storacha/piri/lib/jobqueue"
+	"github.com/storacha/piri/lib/jobqueue/serializer"
+	"github.com/storacha/piri/pkg/pdp/aggregator"
+	"github.com/storacha/piri/pkg/pdp/types"
 )
 
 type Calculator interface {


lib/verifyread/reader.go
```diff
@@ -0,0 +1,71 @@
...
+	src         io.Reader
+	h           hash.Hash
+	expectedSum []byte
+
+	bytesRead uint64
+	done      bool  // reached EOF
+	finalErr  error // latched terminal error (e.g., mismatch)
+}
+
+// If you pass nil values to `New` it will panic and you're an idiot.
```
alanshaw: Return an error instead 😬?

pkg/pdp/piece/resolver.go
```diff
@@ -0,0 +1,121 @@
...
+		// the piece exists in the map, decode it and return
+		read, mh, err := multihash.MHFromBytes(record.Mhash)
+		if err != nil {
+			return nil, false, fmt.Errorf("failed to read multihash: %w", err)
+		}
+		if read != len(record.Mhash) {
+			return nil, false, fmt.Errorf("multihash read mismatch expected %d got %d", len(record.Mhash), read)
+		}
+		return mh, true, nil
+	default: // we resolve the mh to the mh it was uploaded with iff it exists in the store
```
alanshaw: So we're saying if the upload hash was a commp then we don't store a mapping? Wouldn't it just be easier to always store a mapping? - then there's no dependency on the blob store here.

pkg/pdp/httpapi/server/find_piece.go
```diff
@@ -32,11 +31,15 @@ func (p *PDPHandler) handleFindPiece(c echo.Context) error {
 	}
 
 	// Verify that a 'parked_pieces' entry exists for the given 'piece_cid'
-	pieceCID, has, err := p.Service.FindPiece(ctx, types.Piece{
+	mh, err := types.Piece{
 		Name: name,
 		Hash: hash,
 		Size: size,
-	})
+	}.Multihash()
```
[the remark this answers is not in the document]
frrist: Yeah I think I have this wrong, `hash` is already a mutihash so I think I may be double hashing here, which adds an extra prefix, will remove this.

alanshaw (changes requested):

pkg/pdp/piece/commp.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		jobqueue.WithLogger(log.With("queue", CommpQueueName)),
+		// TODO(forrest): determine appropriate amount of retries
+		// failures likely stem from failure to read from blobstore, of invalid cid inputs?
+		jobqueue.WithMaxRetries(50),
+		// TODO(forrest): number of worker will likely be a function of system memory + CPU
+		jobqueue.WithMaxWorkers(uint(runtime.NumCPU())),
+		// TODO(forrest): no idea how long commp will take in practice as it depends on
+		// diskIO for reading, CPU for commp'in and RAM
+		jobqueue.WithMaxTimeout(time.Minute),
```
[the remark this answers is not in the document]
frrist: ehh, this is a confusing option name and my comment is wrong. This value represents the timeout for messages before they can be retried by the worker after a failure. Will just remove this option, the default is 5 seconds.

pkg/pdp/piece/commp.go
```diff
@@ -0,0 +1,128 @@
...
+		&serializer.IPLDCBOR[multihash.Multihash]{
+			Typ:  &schema.TypeBytes{},
+			Opts: captypes.Converters,
+		},
+		jobqueue.WithLogger(log.With("queue", CommpQueueName)),
+		// TODO(forrest): determine appropriate amount of retries
+		// failures likely stem from failure to read from blobstore, of invalid cid inputs?
+		jobqueue.WithMaxRetries(50),
+		// TODO(forrest): number of worker will likely be a function of system memory + CPU
+		jobqueue.WithMaxWorkers(uint(runtime.NumCPU())),
```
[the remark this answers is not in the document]
alanshaw: IDK? `math.Max(1, runtime.NumCPU()-1)`

pkg/pdp/service/piece_resolve.go
```diff
@@ -0,0 +1,23 @@
...
+	return p.pieceResolver.ResolvePiece(ctx, piece)
+}
+
+func (p *PDPService) ResolvePiece(ctx context.Context, piece multihash.Multihash) (_ multihash.Multihash, _ bool, retErr error) {
+	log.Debugw("resolving piece", "request", piece)
+	defer func() {
+		if retErr != nil {
+			log.Errorw("failed to resolve piece", "request", piece, "error", retErr)
+		}
+	}()
```
alanshaw: Why using defer here?

pkg/pdp/service/piece_upload.go
```diff
@@ -42,175 +26,42 @@ func (p *PDPService) UploadPiece(ctx context.Context, pieceUpload types.PieceUpl
...
-	cp := &commp.Calc{}
-	readSize := int64(0)
-
-	var vhash hash.Hash
-	if upload.CheckHashCodec != multihash.Codes[multihash.SHA2_256_TRUNC254_PADDED] {
-		hasher, err := mhreg.GetVariableHasher(multihash.Names[upload.CheckHashCodec], -1)
+	var hasher hash.Hash
+	// TODO where can I get a better value than this shitty string?
+	commpHashCodec := "fr32-sha2-256-trunc254-padded-binary-tree"
+	if upload.CheckHashCodec == commpHashCodec {
```
alanshaw: `doCommp` also allows `multicodec.Sha2_256Trunc254Padded` however, it will be rejected here.

pkg/pdp/service/piece_read.go
```diff
@@ -26,34 +20,12 @@ func (p *PDPService) ReadPiece(ctx context.Context, piece cid.Cid, options ...ty
...
-	var getOptions []blobstore.GetOption
-	if cfg.ByteRange.Start > 0 || cfg.ByteRange.End != nil {
-		getOptions = append(getOptions, blobstore.WithRange(cfg.ByteRange.Start, cfg.ByteRange.End))
-	}
-
-	// TODO(forrest): Nice to have in follow on is attempting to map the `piece` arg to a PieceCIDV2, then
-	// performing the query to blobstore with that CID. allowing the read pieces with the cid they allocated them using
-	obj, err := p.blobstore.Get(ctx, piece.Hash(), getOptions...)
-
+	pr, err := p.pieceReader.ReadPiece(ctx, piece)
```
alanshaw: We need to pass on the options here no?

pkg/service/storage/handlers/blob/accept.go
```diff
@@ -67,36 +63,27 @@ func Accept(ctx context.Context, s AcceptService, req *AcceptRequest) (*AcceptRe
 		}
 	} else {
 		// locate the piece from the pdp service
-		pdpPiece, err := s.PDP().PieceFinder().FindPiece(ctx, req.Blob.Digest, req.Blob.Size)
+		pdpPiece, found, err := s.PDP().API().ResolvePiece(ctx, req.Blob.Digest)
```
alanshaw: We won't be able to accept until commp is calculated if we do this. We should just verify we have it stored as we do above for non-PDP.

pkg/service/storage/handlers/blob/accept.go
```diff
@@ -67,36 +63,27 @@ func Accept(ctx context.Context, s AcceptService, req *AcceptRequest) (*AcceptRe
...
 			log.Errorw("submitting piece for aggregation", "error", err)
 			return nil, fmt.Errorf("submitting piece for aggregation: %w", err)
 		}
-		// generate the invocation that will complete when aggregation is complete and the piece is accepted
-		pieceAccept, err := pdp_cap.Accept.Invoke(
-			s.ID(),
-			s.ID(),
-			s.ID().DID().String(),
-			pdp_cap.AcceptCaveats{
-				Piece: pdpPiece,
```
alanshaw: I would simply change `AcceptCaveats` to use the blob digest. The receipt for this task _already_ communicates the piece CID to the user.

https://github.com/storacha/go-libstoracha/blob/d1d4912dc54d240c54b8f7833d006bf3f3f63cc0/capabilities/pdp/accept.go#L18

alanshaw (changes requested):

pkg/pdp/piece/commp_test.go
```diff
@@ -0,0 +1 @@
+package piece
```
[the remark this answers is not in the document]
frrist: ahh whoops didn't realize this file was empty 🤦 meant to cherry-pick changes from https://github.com/storacha/piri/blob/0e27ec87aafc504cb75f9e3c467c205128212c30/pkg/pdp/service/piece_commp_test.go the PR this was extracted from https://github.com/storacha/piri/pull/328 - will change.

pkg/pdp/service/piece_commp.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return cid.Undef, err
+	}
+	defer piece.Data.Close()
+
+	dmh, err := multihash.Decode(blob)
+	if err != nil {
+		return cid.Undef, fmt.Errorf("invalid multihash: %w", err)
+	}
+	blobCID := cid.NewCidV1(dmh.Code, dmh.Digest)
```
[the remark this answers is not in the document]
frrist: passed multihash, much simpler

pkg/pdp/service/piece_commp.go
```diff
@@ -0,0 +1,167 @@
...
+		// we have a pieceCID v1, convert to v2 and return padded size
+		pieceCID, err := commcid.PieceCidV2FromV1(c, size)
+		if err != nil {
+			return cid.Undef, 0, fmt.Errorf("failed to convert pieceCid %s from v1 to v2: %w", c, err)
+		}
+		treeHeight, _, err := commcid.PayloadSizeToV1TreeHeightAndPadding(size)
+		if err != nil {
+			return cid.Undef, 0, err
+		}
+		return pieceCID, uint64(32) << treeHeight, nil
```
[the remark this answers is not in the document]
frrist: done.

pkg/pdp/service/piece_resolve.go
frrist: removed.

pkg/pdp/service/piece_upload.go
frrist: Implemented a hash registry with a set of supported hash function as you suggested in another comment.

pkg/service/storage/handlers/blob/accept.go
alanshaw: Sorry to be clear, there's no client that reads the data in the caveats, but the data in the receipt for the `pdp/accept` invocation is about to become necessary. Removing the invocation from here makes it harder to lookup the receipt.

We need the inclusion proof in order to sign the "add pieces" message. If we remove this information from the `blob/accept` effects then it's much harder to provide the inclusion proof (it is contained in the receipt for `pdp/accept`).

When asking for an add pieces signing, you're saying you're adding a bunch of roots, and these sub-roots (the blobs) are included in them but there's no proof that they belong to the roots you're adding. i.e. I can't land this change if it's removed: https://github.com/storacha/go-libstoracha/pull/67

Please can we land https://github.com/storacha/go-libstoracha/pull/68 so we don't have to remove?

pkg/pdp/httpapi/client/client.go
```diff
@@ -52,6 +53,38 @@ type Client struct {
...
 
+func (c *Client) ParkPiece(ctx context.Context, params types.ParkPieceRequest) error {
+	return fmt.Errorf("not implemented")
+}
+
+func (c *Client) Has(ctx context.Context, blob multihash.Multihash) (bool, error) {
+	return false, fmt.Errorf("not implemented")
+}
+
+func (c *Client) ResolveToPiece(ctx context.Context, blob multihash.Multihash) (multihash.Multihash, bool, error) {
```
hannahhoward: what is going on here? is this maintained? if not should we drop it?

frrist:
humm, not sure why CI barfs when branches are merged into this, I can't reproduce any of these failures locally... maybe I need to rebase onto main?

PR #334, opened 2025-11-10
feat: forge production presets and preset overhaul

volmedo:
Add presets for forge production services.

The new presets are set as the default because we will want nodes ran with `piri serve` to point to production. This means that, in order to run nodes in the warm staging environment, the command needs to be `PIRI_PRESETS=warm-staging piri serve` (or have the `PIRI_PRESETS` var defined in the environment by any other means).

I did this because I think it is better to screw things in staging because you missed to set the variable than forgetting to make the production presets the default for the production release.

### DIFF
--- a/deploy/full-node/README.md
+++ b/deploy/full-node/README.md
@@ -175,7 +175,7 @@ sudo cat /var/log/cloud-init-output.log
 ## 🔐 SSL Certificates
 
 ### Let's Encrypt Configuration
-- **Production environments** (`environment = "production" or "prod"`): Use Let's Encrypt production certificates (trusted by browsers)
+- **Production environments** (`environment = "production", "prod" or "forge-prod"`): Use Let's Encrypt production certificates (trusted by browsers)
 - **Non-production environments** (staging, dev, etc.): Use Let's Encrypt staging certificates (untrusted, no rate limits)
 
 ### Rate Limits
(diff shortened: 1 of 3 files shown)


pkg/config/pdp.go
```diff
@@ -28,24 +38,65 @@ func (c PDPServiceConfig) Validate() error {
...
+			return app.PDPServiceConfig{}, fmt.Errorf("invalid service address: %s", c.Contracts.Service)
+		}
+		contracts.Service = common.HexToAddress(c.Contracts.Service)
+	}
+	if c.Contracts.ServiceView != "" {
+		if !common.IsHexAddress(c.Contracts.ServiceView) {
+			return app.PDPServiceConfig{}, fmt.Errorf("invalid service view address: %s", c.Contracts.ServiceView)
+		}
+		contracts.ServiceView = common.HexToAddress(c.Contracts.ServiceView)
+	}
```
frrist: blocking: These _must_ all be set, right?

cmd/cli/setup/register.go
frrist: This method returns a config "file" users can consume with `piri serve` - that's how our docs guide them today at least. What do you think of modifying this method to return a config file with all the relevant addresses and urls needed for operation with the `piri serve` command based on the `network` flag provided to the `piri init` command? 

In this way I think we could remove the `network` flag from the `serve` command, and just use the values from the config this command returns. If the values are absent when `piri serve` runs then it can just fail due to an invalid config

frrist (changes requested):
Good stuff, only a couple blocking comments that should be easy enough to address. 

I think we need to consider how the config file returned by the `init` command works with the `serve` command a bit more.

PR #341, opened 2025-11-12
fix: handle errors correctly with middleware

frrist:
Simplify error types and return correct status

### DIFF
--- a/pkg/pdp/httpapi/server/middleware/errors.go
+++ b/pkg/pdp/httpapi/server/middleware/errors.go
@@ -10,90 +10,6 @@ import (
 	"github.com/storacha/piri/pkg/pdp/types"
 )
 
-// ContextualError is a richer error interface that provides additional context
-// about an error that occurred during request handling.
-type ContextualError interface {
-	error
-	// StatusCode returns the HTTP status code that should be returned to the client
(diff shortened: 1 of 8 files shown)


pkg/pdp/httpapi/server/middleware/errors.go
```diff
@@ -105,46 +21,54 @@ func CustomHTTPErrorHandler(err error, c echo.Context) {
...
-		case types.KindConflict:
-			_ = c.String(http.StatusConflict, tErr.Error())
-			return
-		default:
-			_ = c.String(http.StatusInternalServerError, tErr.Error())
-			return
+		if status, ok := typeErrorStatusMap[tErr.Kind()]; ok {
+			return status, tErr.Error()
 		}
+		return http.StatusInternalServerError, tErr.Error()
```
Peeja: **question:** Is `tErr.Error()` better than `err.Error()`? I would expect the outer error to include the inner error's message anyway, and maybe more interesting information too. But I don't know the specifics here.

frrist: The case is basically impossible given the map and type constraint, but need to handle it anyways.

PR #344, opened 2025-11-13
demonstrate bug

frrist:


pkg/service/retrieval/ucan_fx_test.go
```diff
@@ -132,6 +132,72 @@ func TestFXSpaceContentRetrieve(t *testing.T) {
...
+			testutil.Must(
+				delegation.Delegate(
+					space,
+					testutil.Bob,
+					[]ucan.Capability[content.RetrieveCaveats]{
+						ucan.NewCapability(
+							content.RetrieveAbility,
+							space.DID().String(),
+							content.RetrieveCaveats{
+								Blob: content.BlobDigest{Digest: blob.cid.Hash()},
```
frrist: Maybe this isn't a bug, and range is always required to be set? Issue/foot-gun is that when range isn't passed the defaults for `Start` and `End` are `0`, which is valid, but probably not what we want?

PR #348, opened 2025-11-17
feat: send proofs in `SignAddPieces` call

alanshaw:
This PR adds the acceptance store and the receipt store to the PDP service to allow it to retrieve and send required proofs for the `SignAddPieces` call.


hannahhoward (approved):
LGTM yay glad to see this!

PR #390, opened 2025-12-11
feat(service): only calculate piece cids once

hannahhoward:
# Goals

move writing the Piece:MH DB record to commP calc. moreover, make sure that only one calc is in process for a given multihash at any time

# Implementation

- key library is singleflight which I guess is golang stdlib but I didn't know about it. basically allows you to share a function call between go routines so it only gets called for a given set of params once at a time.
- critically, moved writing the PDPPieceCIDToMH record to this function, so that calculation is only ever called once for each multihash (and this will fortunately cause future calls to the piece resolver to succeed)
- wrote a test cause it's a bit complicated, man this whole section is woefully undertested

### DIFF
--- a/pkg/pdp/service/piece_commp_singleflight_test.go
+++ b/pkg/pdp/service/piece_commp_singleflight_test.go
@@ -3,11 +3,13 @@ package service
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"sync"
 	"sync/atomic"
 	"testing"
 	"testing/synctest"
+	"time"
 
 	commcid "github.com/filecoin-project/go-fil-commcid"
 	commp "github.com/filecoin-project/go-fil-commp-hashhash"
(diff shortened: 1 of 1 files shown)


pkg/pdp/service/models/models.go
```diff
@@ -88,6 +88,18 @@ func (ParkedPieceRef) TableName() string {
...
+	Blob       []byte    `gorm:"primaryKey"` // BYTEA primary key
+	StartedAt  time.Time `gorm:"not null"`
+	FinishedAt *time.Time
+	Err        *string `gorm:"column:error"`
+}
+
+func (CommPCalculation) TableName() string {
+	return "commp_calculations"
+}
+
```
frrist: unused?

pkg/pdp/service/piece_commp.go
```diff
@@ -10,26 +10,69 @@ import (
...
+		}
+
+		// 3. insert into pdp_piece_mh_to_commp to avoid recalculation
+		if pieceCID.Hash().HexString() != blob.HexString() {
+			mhToCommp := models.PDPPieceMHToCommp{
+				Mhash: blob,
+				Size:  int64(readObj.Size),
+				Commp: pieceCID.String(),
+			}
+			if err := p.db.Clauses(clause.OnConflict{DoNothing: true}).Create(&mhToCommp).Error; err != nil {
```
frrist: `ModelWithContext` here too I think

alanshaw (approved):
Happy to approve with the caveat that there's clearly some cleanup that needs to be done that @frrist already identified.

PR #400, opened 2025-12-16
telemetry: add deeper metrics to piri

frrist:
Produces the following metrics: 
image

### DIFF
--- a/lib/telemetry/metrics.go
+++ b/lib/telemetry/metrics.go
@@ -137,7 +137,7 @@ func NewTimer(meter metric.Meter, name, description string, boundaries []float64
 }
 
 func (t *Timer) Record(ctx context.Context, duration time.Duration, attrs ...attribute.KeyValue) {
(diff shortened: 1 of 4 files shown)


lib/telemetry/metrics.go
```diff
@@ -0,0 +1,188 @@
+package telemetry
```
hannahhoward: curious why we're implementing all these instead of just using the underlying opentelemetry tools -- some add functionality but others appear to be largely passthroughs

frrist:
> have we verified this all works on staging?

Yes currently deployed to staging, all is well there. 

> have we taken a look at the amount of traffic this generates and verified it isn't super resource intensive?

It should be fairly minimal. Tracing only happens when the client enables it and the number of metrics is fairly minimal for observability, operators may also opt out via an env var.

This change also lays the ground work for an operator to run their own Grafana stack, piri now supports multiple collector endpoints, allows a user to run their own.

Issue #415, opened 2026-01-28
Support for S3 based storage interface


hannahhoward:
Existing "stores"
- Object store
- Acceptance store
- Allocations stores
- Claims stores
- Consolidation 
- Delegation Store
- Receipt Store
- Key Store

1. Evaluate what goes behind an S3 interface
    suggestions:
- key store = no
- evaluate stores 
    -- if the AWS implementation is a Dynamo, probably keep it on disk (and/or put it in the database)
    -- if it's S3 bucket, behind s3 interface

2. Transition relevant stores

- no migration off disk
- if we change any on disk structures to database provide backwards compatibility

PR #417, opened 2026-01-29
Add S3-Compatible Storage Backend

frrist:
closes https://github.com/storacha/piri/issues/415
# Summary
Adds S3-compatible storage as an optional backend while keeping FlatFS as the default. Operators who prefer local filesystems will notice nothing. Operators who want a single node to scale beyond rack limits now have that option.

This implements the Backend Scaling path described in [Piri Storage Scaling](https://www.notion.so/storacha/Scaling-Piri-s-Storage-2ee5305b552480c5a076fe66d2ba95e8). A single Piri node can scale to arbitrary capacity—where "arbitrary" means "whatever your MinIO/Ceph/etc. cluster can handle."

# Motivation
Operators outgrowing Entry stage (60–250 TiB) face a choice: add more nodes (Fleet Scaling) or upgrade to object storage (Backend Scaling). This PR enables the latter.
With S3-compatible storage, operators bring their own backend—MinIO, Ceph, SeaweedFS, etc, and Piri handles the rest. This is configuration, not new architecture.

# Changes

### Storage Module Architecture

Three independent FX storage modules, sharing no imports:

| Module | Location | Purpose |
|--------|----------|---------|
| `filesystem` | `pkg/fx/store/filesystem/` | Local storage via FlatFS and LevelDB |
| `s3` | `pkg/fx/store/s3/` | S3-compatible storage via MinIO client |
| `memory` | `pkg/fx/store/memory/` | In-memory stores for testing |

Backend selection occurs at startup in `pkg/fx/store/provider.go`. Each module is self-contained.

### Bucket Organization
Operators specify a single `bucket_prefix`. Piri creates and manages eight buckets:

| Bucket | Contents | Status |
|--------|----------|--------|
| `{prefix}blobs` | Binary large objects | Currently unused, I think this is leaking from the cloud based deployments |
| `{prefix}allocations` | Space allocations | Required, though could be DB |
| `{prefix}acceptances` | Acceptance records | Required, though could be DB |
| `{prefix}claims` | UCAN claims | Required, though could be DB |
| `{prefix}receipts` | Transaction receipts | Required, though unhappy |
| `{prefix}egress` | Retrieval tracking | Required |
| `{prefix}pdp` | Proof of Data Possession | Required - these are uploads |
| `{prefix}consolidation` | Aggregation state | Required, though unhappy |

### Security Invariant
Private keys never leave local disk. When S3 is configured, KeyStore remains on the filesystem:

```go
// pkg/fx/store/provider.go
return fx.Options(
    s3.Module,                  // All stores except KeyStore
    filesystem.KeyStoreModule,  // KeyStore only—always on disk
)
```
This is not configurable. Some decisions should not be left to operators.

## Configuration

```toml
[repo]
data_dir = '/var/lib/piri'
temp_dir = '/tmp/piri'

[repo.s3]
endpoint = 'minio.example.com:9000'
bucket_prefix = 'piri-'
insecure = false  # true disables TLS (development only)

[repo.s3.credentials]
access_key_id = 'your-access-key'
secret_access_key = 'your-secret-key'
```

When `[repo.s3]` is absent or incomplete, Piri uses the filesystem backend. No migration required for existing deployments.

## Testing

Local verification with MinIO:

```bash
docker run -d \
  --name minio \
  -p 9000:9000 -p 9001:9001 \
  -e MINIO_ROOT_USER=minioadmin \
  -e MINIO_ROOT_PASSWORD=minioadmin \
  minio/minio server /data --console-address ":9001"
```

### Manual Verification Checklist
- [x] Piri starts with S3 configuration
- [x] Eight buckets created with correct prefix
- [x] Blob operations (put/get/delete) work correctly
- [x] KeyStore directory exists on local filesystem
- [ ] Piri falls back to filesystem when S3 config absent
- [ ] Memory stores used when both `data_dir` and S3 are unset

### Minio Console
image

### DIFF
--- a/pkg/store/consolidationstore/legacy.go
+++ b/pkg/store/consolidationstore/legacy.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipfs/go-datastore"
@@ -13,8 +12,6 @@ import (
 
 	"github.com/storacha/piri/pkg/store"
 	"github.com/storacha/piri/pkg/store/consolidationstore/consolidation"
-	"github.com/storacha/piri/pkg/store/objectstore"
-	"github.com/storacha/piri/pkg/store/objectstore/minio"
 )
 
 // Legacy key prefixes for the old two-namespace storage format.
@@ -22,10 +19,6 @@ const (
 	// Datastore prefixes
 	legacyTrackPrefix       = "track/"
 	legacyConsolidatePrefix = "consolidate/"
-
-	// S3/MinIO prefixes
-	legacyS3TrackPrefix       = "consolidation/track/"
-	legacyS3ConsolidatePrefix = "consolidation/consolidate/"
 )
 
 // LegacyReader reads consolidations from the old two-namespace format.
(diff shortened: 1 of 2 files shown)


alanshaw (changes requested):
This is great, but I have the following concerns:

* The stores that have list operations "acceptance" and "allocation" should probably be database tables. Likewise the "receipt" store _index data_ should also be a database table. I believe bucket list operations are slower than database queries, and it ties us to a single index over the data (the key) and prevents the possibility of performing other queries (without scanning the whole dataset).

* I think we should implement FlatFS keys for minio object store like in https://github.com/storacha/piri/pull/408 - I don't know if it's the same for Minio et al. but in S3 there's [guidelines for keys to _not_ all share the same prefix](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-design-patterns.html#optimizing-performance-high-request-rate) in order to optimise request throughput.

  > Why? Because you can achieve 3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD requests per second per partitioned Amazon S3 prefix. So, if all your objects are in the bucket's root they'll all share the same prefix, therefore you can only achieve what I've described before.
  > https://stackoverflow.com/questions/76980883/s3-key-organization-best-practices-uuid-pk-or-name

  Implementing the FlatFS keys by default also allows easier migrations from disk to S3 if necessary.

Overall, I think we _could_ go with this if we really want. It's pleasingly simple, but I don't think that simply because the store was stored on disk it really _should_ be a bucket.

pkg/service/storage/handlers/blob/allocate.go
```diff
@@ -65,23 +65,23 @@ func Allocate(ctx context.Context, s AllocateService, req *AllocateRequest) (res
...
 	}
 
-	allocated := false
-	for _, a := range allocs {
-		if a.Space == req.Space {
-			allocated = true
-			break
-		}
+	// check if any allocation exists for the blob
+	anyAllocation, err := s.Blobs().Allocations().Exists(ctx, req.Blob.Digest)
```
alanshaw: Can probably avoid this call if `s.Blobs().Allocations().Get(ctx, req.Blob.Digest, req.Space)` above returns an allocation.

i.e.

```go
anyAllocation := false
if !allocated {
  exists, _ := s.Blobs().Allocations().Exists(ctx, req.Blob.Digest)
  anyAllocation = exists
}
```

alanshaw (approved):
Ok I see you added the flatfs keys and lets go with S3 for the aformentioned stores. Happy for this to merge when the comments re: allocations are resolved.

PR #419, opened 2026-02-02
fix: allow host and port to be passed to init

frrist:
- allows operators to configure piri server settings from init

### DIFF
--- a/cmd/cli/setup/register.go
+++ b/cmd/cli/setup/register.go
@@ -60,8 +60,8 @@ func init() {
 		"",
 		fmt.Sprintf("Network the node will operate on. This will set default values for service URLs and DIDs and contract addresses. Available values are: %q", presets.AvailableNetworks),
 	)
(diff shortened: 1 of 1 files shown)


Peeja (approved):
I'M YOUR HOST

cmd/cli/setup/register.go
```diff
@@ -60,6 +60,8 @@ func init() {
 		"",
 		fmt.Sprintf("Network the node will operate on. This will set default values for service URLs and DIDs and contract addresses. Available values are: %q", presets.AvailableNetworks),
 	)
+	InitCmd.Flags().String("host", "localhost", "Host piri listens for connections on")
+	InitCmd.Flags().Uint("port", 3000, "Port piri listens for connections on")
```
alanshaw:
```suggestion
	InitCmd.Flags().String("host", "localhost", "Host Piri listens for connections on")
	InitCmd.Flags().Uint("port", 3000, "Port Piri listens for connections on")
```

PR #421, opened 2026-02-04
Payments Settle and Withdraw Implementation w/ TUI 

frrist:


cmd/cli/client/admin/payment/status_model.go
```diff
@@ -0,0 +1,1144 @@
...
+	return func() tea.Msg {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+
+		estimate, err := m.apiClient.EstimateWithdraw(ctx, recipient, "")
+		return withdrawEstimateMsg{estimate: estimate, err: err}
+	}
+}
+
+func (m statusModel) submitWithdraw() tea.Cmd {
```
alanshaw: We should have a separate command to do this as well.

PR #432, opened 2026-02-27
feat: add simple health check to piri api

frrist:


alanshaw (approved):
LGTM but can we remove the z suffix from the endpoints? I'm not sure if there is some significance or just AI being "funny".

PR #467, opened 2026-04-17
Tech Debt Supreme Paydown

frrist:
Removes all legacy code from Piri used for interaction with the Hot Network. :face_exhaling:


alanshaw:
Lets wait to merge this until we have a forked version in the `fil-one` org.

PR #468, opened 2026-04-20
fix: smelt tests

alanshaw:
Updates smelt and fixes up the tests.

Depends on https://github.com/storacha/smelt/pull/15


smelt-tests/smoke_test.go
```diff
@@ -93,7 +89,7 @@ func TestUploadAndRetrieve(t *testing.T) {
 			}
 			t.Log("Added source")
 
-			cids, err := gup.Upload(ctx, spaceDID)
+			cids, err := gup.Upload(ctx, spaceDID, guppy.WithReplicas(1))
```
alanshaw, line 92: ...since there is only one node in the network we need to specify only 1 replica needs to exist.

frrist (approved):

